PR #66, opened 2020-05-07
feat(rns): new placement events

leandroyalet:


src/rns/hooks/utils.ts
```diff
@@ -1,5 +1,17 @@
 import { HookContext } from '@feathersjs/feathers'
 
+export function setSellerAddressParamHook (context: HookContext): HookContext {
```
AuHau, line 3: Sorry about this, my bad. Not sure if we have to have the `utils.ts` file anymore as it was mainly in place to de-duplicate code, but we can keep it for readability I guess.

src/rns/rns.blockchain.ts
```diff
@@ -73,67 +75,85 @@ async function nameChangedHandler (eventData: EventData): Promise<void> {
...
-      throw new Error(`Domain with token ID ${tokenId} not found!`)
-    }
+  const transactionHash = eventData.transactionHash
+  const tokenId = Utils.numberToHex(eventData.returnValues.tokenId)
+
+  const lastOffer = await DomainOffer.findOne({ where: { tokenId: tokenId, status: 'ACTIVE' } })
+
+  if (lastOffer) {
+    logger.info(`Found last offer for ${tokenId}`)
+    lastOffer.status = 'CANCELED'
```
AuHau, line 133: Shouldn't there be different status than `CANCELED` for sold domains?

leandroyalet: have a good eye, I tried to reflect the Placement contract behavior and to avoid duplicate the sold state

AuHau:
Sweet, thanks Leandro!

PR #92, opened 2020-05-12
fix: allows searching own/placed domains by name

jurajpiar:


AuHau:
@jurajpiar did you try to whitelist the `$like` operator instead? As I have already send you... 
https://github.com/feathersjs-ecosystem/feathers-sequelize/blob/master/test/index.test.js#L489

jurajpiar:
Instead of what @AuHau? Whitelisting is not going to help me much in a raw query.

PR #94, opened 2020-05-13
fix: normalized timestamp on old expirationchanged events

leandroyalet:


src/rns/rns.blockchain.ts
```diff
@@ -50,7 +50,12 @@ async function expirationChangedHandler (eventData: EventData): Promise<void> {
   // event ExpirationChanged(uint256 tokenId, uint expirationTime);
 
   const tokenId = Utils.numberToHex(eventData.returnValues.tokenId)
-  const expirationDate = eventData.returnValues.expirationTime * 1000
+  let normalizedTimestamp = eventData.returnValues.expirationTime
+
+  if (normalizedTimestamp.startsWith('10000')) {
+    normalizedTimestamp = eventData.returnValues.expirationTime.slice(5)
+  }
+  const expirationDate = normalizedTimestamp * 1000
```
AuHau, line 58: You should use `parseInt` here because the `normalizedTimestamp` is string and yes you are using the JS "magic" type-conversion here and thanks to that it works, but better to be explicit about it.

PR #102, opened 2020-05-15
fix: includes bought domains in the query

jurajpiar:
After adding the sold status to offers (btw, why not just delete the record?) the query was not including those records for the users that bought the domain.
Adding to the pile, this extends the raw sql query to also include the 'SOLD' offers in the owned domains query.

Disclaimer: Although the query will have to be cleaned up, rather sooner than later, at the moment the priority lays with getting it work. Refactor will come later.

### DIFF
--- a/src/rns/hooks/domain.hooks.ts
+++ b/src/rns/hooks/domain.hooks.ts
@@ -65,18 +65,21 @@ export default {
                   FROM
                       "rns_domain-offer"
                   WHERE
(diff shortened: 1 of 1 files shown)


AuHau:
This is a bit question for @leandroyalet but I am also not sure why we need to have the SOLD state? Or actually any state? Shouldn't it be the case for `DomainOffer` table to have only current offers? Why should we have CANCELED and SOLD as well?

PR #117, opened 2020-05-21
feat: store cache-state data in database

### LINKED ISSUE #115: Drop conf package
Currently Cache uses [`conf`](https://github.com/sindresorhus/conf) package for storing some data (last processed blocks) outside of database, which might be confusing for people using this project (for example they will truncate database and the data won't be re-loaded because the Cache will use the last-processed-block from conf which will be different the `null`)

All data should be stored in DB.

AuHau:
Closes #115 

Will require everybody to delete database and run `precache` command.


julianmrodri:
@AuHau is this always going to be the case to initial setup to run Precache and then Start?

AuHau:
Well, not exactly sure what you mean. If you are asking about the comment I put to the description it is meant as a reminder to tell everybody to make this (eq. delete DB and precache) after they pull these changes, otherwise they might get weird behavior from Cache...

src/rns/index.ts
```diff
@@ -139,9 +141,11 @@ const rns: CachedService = {
     const domainsCount = await Domain.destroy({ where: {}, truncate: true, cascade: true })
     logger.info(`Removed ${offersCount} offers entries, ${soldCount} sold domains and ${domainsCount} domains`)
 
-    confFactory().delete('rns.owner')
-    confFactory().delete('rns.reverse')
-    confFactory().delete('rns.placement')
+    const store = getObject()
+    delete store['rns.placement.lastProcessedBlock']
+    delete store['rns.reverse.lastProcessedBlock']
+    delete store['rns.reverse.lastProcessedBlock']
```
leandroyalet, line 147: this line is repeated

AuHau: Oh! Good catch! Copy&paste 😅 Thanks 😊

julianmrodri:
> Well, not exactly sure what you mean. If you are asking about the comment I put to the description it is meant as a reminder to tell everybody to make this (eq. delete DB and precache) after they pull these changes, otherwise they might get weird behavior from Cache...

Adam what I meant is that started everything from scratch and couldnt run Start as before. It forced me to execute Precache and then Start as two separate steps to run the Cache.

PR #125, opened 2020-05-26
Feat/confirmation endpoint

AuHau:


src/rns/index.ts
```diff
@@ -61,7 +61,7 @@ async function precache (eth?: Eth): Promise<void> {
   await fetchEventsForService(eth, 'rns.placement', simplePlacementsContractAbi as AbiItem[], dataQueuePusher)
 
   // We need to sort the events in order to have valid
-  eventsDataQueue.sort((a, b): number => {
+  const sortedQueue = eventsDataQueue.sort((a, b): number => {
```
vojtechsimetka, line 64: This assignment is unnecessary, they both point to the same array as `Array.sort` is in place and modifies original array https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

src/rns/index.ts
```diff
@@ -61,7 +61,7 @@ async function precache (eth?: Eth): Promise<void> {
...
     // First by block number
     if (a.blockNumber - b.blockNumber !== 0) return a.blockNumber - b.blockNumber
```
vojtechsimetka, line 66: I would prefer this to be:

`if (a.blockNumber !== b.blockNumber) return a.blockNumber - b.blockNumber`

vojtechsimetka:
Also please fix that unused import `asyncSplit` https://lgtm.com/projects/g/rsksmart/rif-marketplace-cache/rev/pr-0d39e89e1f5abf356392d2f2e87f389dc5c02c9c

src/rns/index.ts
AuHau: Right, I thought it is in place, but when I was checking it I just saw the "Return value = sorted array" and did not finish reading to verify that it was in place 🙈 Thanks for the catch

PR #130, opened 2020-05-28
feat: uses services for db interaction

### LINKED ISSUE #51: Cache does not fire events (created, updated).
This is likely caused by calling the create, update methods on the `Model` object instead of `Service`.
According to the feathers docs, ["Any service automatically emits created, updated, patched and removed events "](https://docs.feathersjs.com/api/events.html#service-events).
From their examples can be observed that the create method are called upon the service object not the model.
```
const feathers = require('@feathersjs/feathers');
const app = feathers();

app.use('/messages', {
  create(data, params) {
    return Promise.resolve(data);
  }
});

// Retrieve the wrapped service object which will be an event emitter
const messages = app.service('messages');

messages.on('created', (message, context) => console.log('created', message));

messages.create({
  text: 'We have to do something!'
});

```
One way to fix this is to rewrite the rns.blockchain.ts file to use the service object for creation/update of data.
Another, more hacky way would be to create [custom events](https://docs.feathersjs.com/api/events.html#custom-events).

#### Fixed in

 - [ ] Storage
 - [ ] RNS
 - [ ] Rates 

Blocking [Capture events from the cache service #51](https://github.com/rsksmart/rif-marketplace-ui/issues/51)

jurajpiar:
Closes #51 
Relates to [UI#155](https://github.com/rsksmart/rif-marketplace-ui/pull/155)


src/rns/index.ts
```diff
@@ -89,11 +89,28 @@ const rns: CachedService = {
...
+    app.use('/rns/v0/offers', new RnsService({ Model: DomainOffer, multi: true }))
+
+    const domains = app.service('/rns/v0/:ownerAddress/domains')
+    domains.hooks(domainHooks)
+    const sold = app.service('/rns/v0/:ownerAddress/sold')
+    sold.hooks(soldDomainHooks)
+    const offers = app.service('/rns/v0/offers')
+    offers.hooks(domainOfferHooks)
+
+    app.on('connection', (connection: any) => {
```
AuHau, line 101: Interesting. Is it really needed though? 
I guess what are you doing here that you are creating special channels for the updates right? I guess this is why you said that you will remove the `:ownerAddress` right? After that change this won't be needed, because you will have the automatic channel like `/rns/v0/offers` that you can listen on directly, right?

src/rns/rns.processor.ts
```diff
@@ -178,15 +192,15 @@ const commands = {
...
   return value in commands
 }
 
-export default function rnsProcessorFactory (logger: Logger, eth: Eth) {
+export default function rnsProcessorFactory(logger: Logger, eth: Eth, app?: Application) {
   return async function (eventData: EventData): Promise<void> {
     if (isValidEvent(eventData.event)) {
       logger.info(`Processing event ${eventData.event}`)
-      await commands[eventData.event](logger, eventData, eth)
+      await commands[eventData.event](logger, eventData, eth, app)
```
AuHau, line 203: Watch out, you have `app` defined as `?` but in the handlers, you use it without the `?` (eq. you don't check if it exists). Fix that, please.

Also, think about how this will work for the `precache` CLI command. To be honest I don't like passing the whole `app` to the processor. Somehow I  would like to keep them separated for the "separation of concerns". The processor should not have to know on what routes are the services defined etc.

What about having an object like this:
```
const services = {domains, offers, sold}
```
and pass that instead of the `app`?

src/rns/index.ts
jurajpiar: Yea I thought that would be so too (as that's how it is described in the docs), but in reality event the offers connection has not been publishing events. Thus I created a channel for that endpoint too.

AuHau: Hmmm that sounds suspicious 🤔 Did you do some more digging/research on that?

src/rns/rns.processor.ts
jurajpiar: sure, that can be done. on the other hand, having the entire context available seems to be the feathers way. That is why I've done it so. But I have no quarrel with passing just the services.
As to the precache, it will not use that, that is why it is optional in the factory. And yes, it does need to check whether it exists and fallback to using the model.

AuHau: Well, one (me :-P) could argue that Cache consists of two parts blockchain layer listening **and processing** events from blockchain and then presentation layer that is built with FeathersJs. These two layers are "connected" through DB (with introducing this change this becomes less true, unfortunately), but generally these two layers should be kept separate.

Also, you are wrong that `precache` won't use it. It will ;-) It uses the same processor and also needs a way how to store the the data...

src/services/rns/rns.processor.ts
```diff
@@ -37,25 +33,28 @@ async function transferHandler (logger: Logger, eventData: EventData): Promise<v
...
   if (transferDomain) {
     logger.info(`Transfer event: Transfer ${tokenId} created`)
   }
+
+  const domainsService = services.domains
   const domain = await Domain.findByPk(tokenId)
 
   if (domain) {
     await DomainOwner.upsert({ address: ownerAddress, tokenId })
+    await domainsService.patch(domain.tokenId, { tokenId }) // Trigger patch event
```
AuHau, line 48: Wait... 😳 So you are updating here `tokenId` to the same value, just to trigger the event?

Did you try to use simple `domainsService.emit('patched', <data>)` instead?

src/services/rns/rns.processor.ts
```diff
@@ -65,15 +64,20 @@ async function expirationChangedHandler (logger: Logger, eventData: EventData):
...
 
   if (currentExpiration) {
     await DomainExpiration.update({ expirationDate }, { where: { tokenId } })
+    domainsService.patch(tokenId, { tokenId }) // Trigger patch event
     logger.info(`ExpirationChange event: DomainExpiration for token ${tokenId} updated`)
   } else {
-    await Domain.upsert({ tokenId })
+    const domain = await Domain.findByPk(tokenId)
+
+    if (!domain) domainsService.create({ tokenId })
```
AuHau, line 80:
```suggestion
    if (!domain) await domainsService.create({ tokenId })
```

src/services/rns/rns.processor.ts
```diff
@@ -82,25 +86,25 @@ async function expirationChangedHandler (logger: Logger, eventData: EventData):
...
 
   const label = name.substring(0, name.indexOf('.'))
   const tokenId = Utils.numberToHex(Utils.sha3(label) as string)
 
   const domain = await Domain.findByPk(tokenId)
 
   if (domain) {
-    domain.name = name
-    domain.save()
+    domainsService.update(tokenId, { name })
```
AuHau, line 99:
```suggestion
    await domainsService.update(tokenId, { name })
```

src/services/rns/rns.processor.ts
```diff
@@ -82,25 +86,25 @@ async function expirationChangedHandler (logger: Logger, eventData: EventData):
...
     logger.info(`NameChanged event: Updated Domain name ${name} -> ${tokenId}`)
   } else {
-    await Domain.create({ tokenId, name })
+    domainsService.create({ tokenId, name })
```
AuHau, line 102:
```suggestion
    await domainsService.create({ tokenId, name })
```

src/services/rns/rns.processor.ts
```diff
@@ -113,54 +117,55 @@ async function tokenPlacedHandler (logger: Logger, eventData: EventData, eth: Et
...
+async function tokenUnplacedHandler (logger: Logger, eventData: EventData, eth: Eth, services: RnsServices): Promise<void> {
   // event TokenUnplaced(uint256 indexed tokenId);
-
   const tokenId = Utils.numberToHex(eventData.returnValues.tokenId)
-
   const storedOffer = await DomainOffer.findOne({ where: { tokenId } })
 
   if (storedOffer) {
-    storedOffer.destroy()
+    const offersService = services.offers
```
AuHau, line 149: You can directly destructuralize the `services` objects into the services you need in the function declaration... I think that would be nicer 😇
(applicable to the whole processor)

src/services/rns/hooks/domain.hooks.ts
```diff
@@ -39,6 +39,8 @@ export default {
         if (context.params.query) {
           const { ownerAddress, placed, name } = context.params.query
 
+          if (!ownerAddress) { return Promise.reject(new Error('No ownerAddress specified.')) }
```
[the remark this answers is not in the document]
jurajpiar: Ah yea, not sure. Will change.

src/services/rns/hooks/sold-domain.hooks.ts
```diff
@@ -52,14 +57,14 @@ export default {
           }
         }
 
-        discardQuery('domain')
+        context.params.query = {}
```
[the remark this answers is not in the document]
jurajpiar: Yea I know. There isn't anything else we filter on this table other than what is already used in the query, that's why I cleared the whole *context.params.query*. But we might in the future so I will discard the params separately.

src/services/rns/index.ts
```diff
@@ -19,18 +20,23 @@ import domainHooks from './hooks/domain.hooks'
...
 
-import rnsContractAbi from '@rsksmart/rns-rskregistrar/RSKOwnerData.json'
-import rnsReverseContractAbi from '@rsksmart/rns-reverse/NameResolverData.json'
-import auctionRegistrarContractAbi from '@rsksmart/rns-auction-registrar/TokenRegistrarData.json'
-import simplePlacementsContractAbi from '@rsksmart/rif-marketplace-nfts/ERC721SimplePlacementsABI.json'
 import { errorHandler, waitForReadyApp } from '../../utils'
 
 import { processRskOwner, processAuctionRegistrar } from './rns.precache'
+import rnsChannels from './rns.channels'
+import Eth from 'web3-eth'
```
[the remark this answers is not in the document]
jurajpiar: Ah cool, didn't know about the *import type* syntax.
The order is a no-brainer, of course.

src/services/rns/rns.processor.ts
jurajpiar: Not sure why I did that, lol.

jurajpiar: Ah no I do know. because we don't always have access to the emit object. but I guess I could just check instead of writing to db unnecessarily

src/services/rns/rns.processor.ts
jurajpiar: that was a choice.
it wouldn't be nicer as it would have to be either:
`const { offers } = services`
in which case it's unclear what the constant is, or:
`const { offers: offersService } = services`
which does not even need a comment

PR #140, opened 2020-06-03
feat: add entrypoint for running required steps

lucasvuotto:
While there, adjust DEPLOYMENT.md accordingly. Also add an environment
variable in Dockerfile to allow specifying specifics targets for
`precache` subcommand.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,9 @@ RUN sed -i 's#"./src/cli"#"./lib/cli"#g' package.json
 
 EXPOSE 3030
 
-ENTRYPOINT [ "./bin/run" ]
-CMD [ "start" ]
+ENV RIFM_PRECACHE_TARGETS all
+
+ENTRYPOINT [ "./bin/entrypoint" ]
 
 LABEL maintainer="adam@iovlabs.org"
 LABEL description="Blockchain caching server for RIF Marketplace"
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -26,8 +26,9 @@ RUN sed -i 's#"./src/cli"#"./lib/cli"#g' package.json
 
 EXPOSE 3030
 
-ENTRYPOINT [ "./bin/run" ]
-CMD [ "start" ]
+ENV RIFM_PRECACHE_TARGETS rns rates
```
AuHau, line 29: Hmmm I am pretty sure that `ENV` is only for build-time and not run-time. So I would suggest removing this. Btw. there is also an option "all" which will precache all the services, which I would make a default if the `RIFM_PRECACHE_TARGETS` is not set.

AuHau:
I will tackle the linting problem.

AuHau:
@lucasvuotto I have fixed the linting in the `master` so please rebase and it should not bother you anymore ;-)

Dockerfile
lucasvuotto: From https://docs.docker.com/engine/reference/builder/#env

> The environment variables set using `ENV` will persist when a container is run from the resulting image. You can view the values using `docker inspect`, and change them using `docker run --env <key>=<value>`.

Didn't know about `all`. Will use it instead.

lucasvuotto:
Thanks for that! Rebased.

PR #141, opened 2020-06-04
feat: adds expiration table and removes status from offers

### LINKED ISSUE #93: Remove SQL from domain.hooks.ts
As part of temporal workaround for the presentation there were some code-quality compromises that should be revisited later on. Part of it is to remove custom SQL query from https://github.com/rsksmart/rif-marketplace-cache/blob/master/src/rns/hooks/domain.hooks.ts

### LINKED ISSUE #131: Remove "Status" from all queries
Remove Status from all Queries in RNS But. There is a place where we cancel past offers. Instead of cancel we need to do update the existing one and put the create new on in the else statement. (or delete the existing and always create).

jurajpiar:
Closes #131
Closes #93 
Plus extracts expiration date from domains in order to prevent racing condition on blockchain transfer events.
Makes a tiny refactor of services' paths in the definitions and rns.

Sold process doesn't seem to work yet (though triggered) and transfer events seem to be only triggered in precache, thus the PR is a draft.
Otherwise seems to be working fine.

### DIFF
--- a/src/services/rns/hooks/domain.hooks.ts
+++ b/src/services/rns/hooks/domain.hooks.ts
@@ -1,5 +1,5 @@
 import { HookContext } from '@feathersjs/feathers'
-import { disallow } from 'feathers-hooks-common'
+import { disallow, discardQuery } from 'feathers-hooks-common'
 import { Op } from 'sequelize'
 import { numberToHex, sha3 } from 'web3-utils'
 import DomainOffer from '../models/domain-offer.model'
@@ -74,8 +74,8 @@ export default {
               }
             }
           }
-          delete (context.params.query as any).status
-          delete (context.params.query as any).name
+          discardQuery('placed')
+          discardQuery('name')
         }
       }
     ],
--- a/src/services/rns/hooks/domain.hooks.ts
+++ b/src/services/rns/hooks/domain.hooks.ts
@@ -1,8 +1,9 @@
 import { HookContext } from '@feathersjs/feathers'
 import { disallow } from 'feathers-hooks-common'
-import { sha3, numberToHex } from 'web3-utils'
-import Domain from '../models/domain.model'
+import { Op } from 'sequelize'
+import { numberToHex, sha3 } from 'web3-utils'
 import DomainOffer from '../models/domain-offer.model'
+import DomainExpiration from '../models/expiration.model'
 
 export default {
   before: {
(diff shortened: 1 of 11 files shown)


src/services/rns/models/expiration.model.ts
```diff
@@ -0,0 +1,12 @@
+import { Column, DataType, Model, Table, HasMany, BelongsTo, ForeignKey } from 'sequelize-typescript'
+import Domain from './domain.model'
+
+
+@Table({ freezeTableName: true, tableName: 'rns_domain_expiration', timestamps: false })
+export default class DomainExpiration extends Model {
```
AuHau, line 6: Hmmmm 😅

Is this new Model the solution for the "synchronization problem"?

Because from the database point of view I don't like this design in the same way as I didn't like the `OwnerAddress` model proposition you had earlier 😅

src/services/rns/rns.processor.ts
```diff
@@ -66,12 +57,22 @@ async function expirationChangedHandler (logger: Logger, eventData: EventData):
     normalizedTimestamp = eventData.returnValues.expirationTime.slice(5)
   }
   const expirationDate = parseInt(normalizedTimestamp) * 1000
-  await Domain.upsert({ tokenId, expirationDate })
 
-  logger.info(`ExpirationChange event: Domain ${tokenId} updated`)
+  const currentRecord = await DomainExpiration.findOne({ where: { tokenId } })
```
AuHau, line 61: Out of curiosity any reason why not to use the `upsert()` here?

src/services/rns/models/expiration.model.ts
AuHau: Hmm re-read the PR's description, so found an answer to the question.

I see, alright :-)

src/services/rns/hooks/domain.hooks.ts
```diff
@@ -34,73 +35,117 @@ export default {
         return context
       },
       async (context: HookContext) => {
-        if (context.params.query?.status) {
-          const { status, ownerAddress, name: nameFilter } = context.params.query
+        if (context.params.query) {
+          const { ownerAddress, placed, name } = context.params.query
+          // @ts-expect-error
```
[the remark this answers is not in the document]
jurajpiar: Oh that was there temporarily and I forgot to remove it. thanks

src/services/rns/rns.processor.ts
```diff
@@ -9,54 +9,45 @@ import Utils from 'web3-utils'
 import { Eth } from 'web3-eth'
 import { getBlockDate } from '../../blockchain/utils'
 import { Logger } from '../../definitions'
+import DomainExpiration from './models/expiration.model'
+import config from 'config'
 
-async function transferHandler (logger: Logger, eventData: EventData): Promise<void> {
```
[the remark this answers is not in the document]
jurajpiar: Not sure what you are referring to here. Is it the space? that's my formatter. I guess I should fix it because it adds the space everywhere I visit making PRs less clear.

src/services/rns/rns.processor.ts
jurajpiar: I just wanted to log when it finds one, which upsert doesnt tell you. But that doesn't matter much as I'll be converting all of this to using the services in the PR that will follow this one.

src/services/rns/rns.processor.ts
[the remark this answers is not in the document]
AuHau: Yeah the space :-) 
Sounds good :-)

src/services/rns/rns.processor.ts
AuHau: Make sense 👍

src/services/rns/hooks/domain.hooks.ts
```diff
@@ -33,74 +34,49 @@ export default {
...
+                  [Op.like]: `%${nameFilter}%`
+                },
+                tokenId: {
+                  [Op.eq]: numberToHex(((sha3(nameFilter)) as string))
+                }
               }
             }
-          })
+          }
+          delete (context.params.query as any).status
```
AuHau, line 77: I was looking around how to do this better as I don't like the cast to `any` and discovered https://hooks-common.feathersjs.com/hooks.html#discardquery

I would suggest using this.

jurajpiar: ah, nice found!

PR #145, opened 2020-06-07
fix: makes precache update expiration before domain table

jurajpiar:
adds id to expiration table definition

As I had a different version of the ui-config that was generated by the dev project, I was missing the auction registrar address and thus did not capture the error in my last PR.
This is to fix the issue with foreign key constraints when writing to domains table.

### DIFF
--- a/src/services/rns/hooks/domain.hooks.ts
+++ b/src/services/rns/hooks/domain.hooks.ts
@@ -1,9 +1,10 @@
 import { HookContext } from '@feathersjs/feathers'
-import { disallow, discardQuery } from 'feathers-hooks-common'
+import { disallow } from 'feathers-hooks-common'
 import { Op } from 'sequelize'
 import { numberToHex, sha3 } from 'web3-utils'
 import DomainOffer from '../models/domain-offer.model'
 import DomainExpiration from '../models/expiration.model'
+import DomainOwner from '../models/owner.model'
 
 export default {
   before: {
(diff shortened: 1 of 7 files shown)


julianmrodri:
@jurajpiar  I tried this and get Invalid Date in the UI for all domains. is this related? Maybe the Expired date is not being stored properly? 
image

src/services/rns/models/expiration.model.ts
```diff
@@ -3,6 +3,9 @@ import { Column, DataType, Model, Table } from 'sequelize-typescript'
 @Table({ freezeTableName: true, tableName: 'rns_domain_expiration', timestamps: false })
 export default class DomainExpiration extends Model {
   @Column({ primaryKey: true, type: DataType.STRING })
+  id!: string
+
+  @Column({ type: DataType.STRING })
   tokenId!: string
```
vojtechsimetka, line 9: Do I understand it right, that you are duplicating the `tokenId`? This seems like quite lazy solution.

src/services/rns/hooks/domain.hooks.ts
```diff
@@ -74,8 +82,9 @@ export default {
               }
             }
           }
-          discardQuery('placed')
-          discardQuery('name')
+          // discardQuery('placed')
```
AuHau, line 85: Remove this, please :-)

julianmrodri:
Worked fine. Lets correct the suggestions and get this merged!

src/services/rns/hooks/domain.hooks.ts
jurajpiar: oh yea thanks

PR #146, opened 2020-06-07
feat: added config for default deterministic ganache

vojtechsimetka:


julianmrodri:
Nice! this would remove the need to copy the config file from the Dev project.

julianmrodri:
Shouldnt we add "host" (localhost) and "port" (3030)?

vojtechsimetka:
@julianmrodri the only thing needed is port and I added it. Now running `NODE_ENV=ganache npm run start` should work out of the box.

PR #155, opened 2020-06-09
feat: add confirmations to RNS contracts

julianmrodri:
Adding  default confirmations = 2 for  RNS contracts in `ganache.json5`.

### DIFF
--- a/config/ganache.json5
+++ b/config/ganache.json5
@@ -11,31 +11,31 @@
     owner: {
       contractAddress: "0x4bf749ec68270027C5910220CEAB30Cc284c7BA2",
       eventsEmitter: {
-        confirmations: 0
+        confirmations: 2
       }
     },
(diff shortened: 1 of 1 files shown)


config/ganache.json5
```diff
@@ -5,6 +5,7 @@
   },
   storage: {
     enabled: false,
```
AuHau, line 7: Well you might want to enable this 😉 Otherwise the address won't be much to use...

nduchak: Sorry guys, I resolve the merge conflicts by mistake. Think it was my PR

julianmrodri:
Hi, because the Storage service is not ready to work with Cache i used this branch and PR for the Confirmations number when running Ganache in deterministic mode, which is something we need for RNS.

PR #158, opened 2020-06-12
feat: confirmations improvements

### LINKED ISSUE #157: Confirmations improvements
After talking with Raulph we indentified few short-comings of our current Confirmations process. With @vojtechsimetka we have designed few improvements that needs to be incorporated.

AuHau:
Closes #157


src/blockchain/new-block-emitters.ts
```diff
@@ -0,0 +1,129 @@
...
+export abstract class AutoStartStopEventEmitter extends EventEmitter {
+  /**
+   * Name of event that triggers the start/stop actions. Eq. waits there is listeners for this specific event.
+   */
+  private readonly triggerEventName: string
+  protected logger: Logger
+
+  protected constructor (logger: Logger, triggerEventName: string) {
+    super()
+    this.logger = logger
```
vojtechsimetka, line 23: Do you need to bind the logger? Can it not live as a global variable or is the idea that the `start` and `stop` handlers can access it from `this`? Did you consider passing it as argument instead?

src/utils.ts
```diff
@@ -121,3 +121,11 @@ export abstract class BaseCLICommand extends Command {
     return Promise.resolve()
   }
 }
+
+export function setDifference<T> (setA: Set<T>, setB: Set<T>): Set<T> {
```
vojtechsimetka, line 125: Isn’t this rather `setExtent`?

config/default.json5
```diff
@@ -11,21 +11,11 @@
...
-      newBlockEmitter: {
-        // If to use polling strategy, if false then listening is used.
-        polling: true
-      }
-    }
+    // Multiplier that is used for targetConfirmations that determines when an event
+    // from DB is supposed to be removed.
+    // Eq. if event is supposed to be confirmed after 5 blocks (eq. targetConfirmations)
+    // when this parameter is set to "1.5" then the event will be removed after 8 confirmations.
+    deleteTargetConfirmationsMultiplier: 1.5
```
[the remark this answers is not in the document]
AuHau: After I have explained the multiplier to you, I realized that this is not so good approach as it is not "deterministic enough" and hard to understand. I will refactor it to more clear `waitBlockCountToDelete` approach, eq. specifing number of blocks before it is removed from DB.

src/blockchain/new-block-emitters.ts
AuHau: This is because it is a base class for multiple different classes and you would get very unclear logs. Eq. you would not be able to distinguish what classes started/stopped listening.

src/utils.ts
AuHau: No. This is copy&pasted from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set 😅

PR #159, opened 2020-06-14
feat: adds name resolution to the rns processor

jurajpiar:


julianmrodri:
Domains are showing up without the TLD. Are we loosing it somewhere in the Decoding?
image

julianmrodri:
Also registered a 2nd domain and never showed up.

PR #175, opened 2020-06-21
fix(core): confirmations are handled for multiple blocks

AuHau:
If there are processed mutiple blocks in one confirmation invokation
some events could be skipped over.

### DIFF
--- a/src/blockchain/confirmator.ts
+++ b/src/blockchain/confirmator.ts
@@ -12,7 +12,7 @@ import type { EventEmitter } from 'events'
 import { literal, Op } from 'sequelize'
 
 function isConfirmedClosure (currentBlockNumber: number) {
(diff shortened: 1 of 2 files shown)


src/blockchain/confirmator.ts
```diff
@@ -12,7 +12,7 @@ import type { EventEmitter } from 'events'
 import { literal, Op } from 'sequelize'
 
 function isConfirmedClosure (currentBlockNumber: number) {
-  return (event: Event): boolean => event.getConfirmationsCount(currentBlockNumber) === event.targetConfirmation
+  return (event: Event): boolean => event.getConfirmationsCount(currentBlockNumber) >= event.targetConfirmation
```
jurajpiar, line 15: :D

Issue #139, opened 2020-06-02
Refactor Storage processor to use Services


AuHau:

AuHau:
Blocked by #130

AuHau:
Closed by #182

PR #182, opened 2020-06-22
Feat/storage services

### LINKED ISSUE #46: Track PeerID
When new Offer is created, the Provider can emit a Message with metadata. Currently the most important is PeerID of the its IPFS node. This allows for the Consumer to directly check with the IPFS node of the Provider in order to verify that the file was pinned and is provided by him.

AuHau:
Closes #46 and #139


src/services/storage/index.ts
```diff
@@ -5,19 +5,31 @@ import { getObject } from 'sequelize-store'
...
 export class OfferService extends Service {
+  emit?: Function
+}
+
+export class AgreementService extends Service {
+  emit?: Function
+}
+
+export interface StorageServices {
+  agreementService: AgreementService & { emit?: Function }
```
jurajpiar, line 31: is this `AgreementService & { emit?: Function }` necessary if we have the emit in the class definition? Also wouldn't it be better to have something like
```
export interface StorageService extends Service {
  emit?: Function
}

export interface StorageServices {
  agreementService: StorageService,
  offerService: StorageService
}
```
alternatively (should they ever differ)
```
class AgreementService implements StorageService { ... }
...
```?

Comment (jurajpiar) [reply src/services/storage/index.ts:31]: but I guess it's not a biggie anyway so feel free to ignore it. It's just that I would personally find that clearer.

File: src/utils.ts [hunk 2]
```diff
@@ -23,16 +23,60 @@ export async function asyncSplit<T> (arr: T[], callback: (elem: T) => Promise<bo
...
+/**
+ * Utility function for decoding Solidity's byte32 array.
+ * @param fileReference
+ */
+export function decodeByteArray (fileReference: string[]): string {
+  return fileReference
+    .map(hexToAscii)
+    .join('')
+    .trim()
+    .replace(/\0/g, '') // Remove null-characters
```
jurajpiar, line 47: I'm not trying to question the decision here but it poked me in the eye as I don't understand the reason for it. Do we really want to delete all the zeroes from the ASCII representation?

src/services/storage/hooks/offers.hooks.ts
```diff
@@ -29,7 +31,7 @@ export default {
   },
 
   after: {
-    all: [],
+    all: [dehydrate(), discard('agreements')],
```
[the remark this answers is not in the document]
AuHau: Yes, because we use `raw: false` and the `feather-common-hooks` don't really work with the ORM instances.

src/services/storage/index.ts
AuHau: I copy-pasted from RNS (a.k.a yours) part 😝  But yeah agree, if it will work I am down.

src/utils.ts
AuHau: It is not a zero but a null-characters as the comment says ;-) They are represented as `\0`. Fore removing zeros it would be without the `\`.

jurajpiar: ah yea... oops

src/services/storage/index.ts
jurajpiar: :D

src/services/storage/handlers/agreement.ts
```diff
@@ -106,14 +109,14 @@ function isValidEvent (value: string): value is keyof typeof handlers {
...
-  handler (event: EventData, eth: Eth): Promise<void> {
+  handler (event: EventData, services: StorageServices, eth: Eth): Promise<void> {
     if (!isValidEvent(event.event)) {
       return Promise.reject(new Error(`Unknown event ${event.event}`))
     }
 
-    return handlers[event.event](event, eth)
+    return handlers[event.event](event, services, eth)
   }
 }
```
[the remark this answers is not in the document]
AuHau: Good point, I guess best option is to find different name for the `handler` function, which btw. should have been originally called `handle`  😅

src/services/rns/rns.channels.ts
```diff
@@ -7,11 +7,11 @@ export default function (app: Application) {
     return
   }
   app.on('connection', (connection: any) => {
-    app.channel('offers').join(connection)
-    app.channel('domains').join(connection)
-    app.channel('sold').join(connection)
+    app.channel('rns_offers').join(connection)
+    app.channel('rns_domains').join(connection)
+    app.channel('rns_sold').join(connection)
```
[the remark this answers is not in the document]
AuHau: Agree, I will drop it for now, but yeah I have realized it is good idea to prefix the channels as for example the storage has also "offers" channels ;-)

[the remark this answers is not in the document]
jurajpiar: yes, I agree

PR #184, opened 2020-06-23
feat: wrapps storing name in the precache in try

jurajpiar:
Fixes issue on a testnet which seems to fail the process of precache on a strange name resolution.

### DIFF
--- a/src/services/rns/rns.precache.ts
+++ b/src/services/rns/rns.precache.ts
@@ -57,7 +57,12 @@ export async function processRskOwner (eth: Eth, logger: Logger, contractAbi: Ut
     const name = Utils.hexToAscii('0x' + decodedData.params[2].value.slice(218, decodedData.params[2].value.length))
     const tokenId = Utils.numberToHex(Utils.sha3(name) as string)
     const ownerAddress = rskOwnerEvent.returnValues.to.toLowerCase()
-    await Domain.upsert({ tokenId, name: `${name}.${tld}` })
+    try {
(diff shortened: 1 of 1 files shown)


src/services/rns/rns.precache.ts
```diff
@@ -57,7 +57,11 @@ export async function processRskOwner (eth: Eth, logger: Logger, contractAbi: Ut
     const name = Utils.hexToAscii('0x' + decodedData.params[2].value.slice(218, decodedData.params[2].value.length))
     const tokenId = Utils.numberToHex(Utils.sha3(name) as string)
     const ownerAddress = rskOwnerEvent.returnValues.to.toLowerCase()
-    await Domain.upsert({ tokenId, name: `${name}.${tld}` })
+    try {
+      await Domain.upsert({ tokenId, name: `${name}.${tld}` })
+    } catch (e) {
+      logger.warn(`Domain name ${name}.${tld} for token ${tokenId} could not be stored.`)
```
julianmrodri, line 63: I think we should UPSERT with null or empty name, so we dont lose the relationship. Not just log the issue.

julianmrodri: I think we are using NULL for domains that we could not retrieve the"readable" name

AuHau:
@jurajpiar I would argue that the commit message should have had "fix" scope (eq. `fix:`)... Just for future...

jurajpiar:
@AuHau yes you are right

PR #191, opened 2020-06-26
feat(storage): unit test's

nduchak:
- Provide fix for `Offer` model virtual field
 - Provide fix for message parsing in `MessageEmitted` event handler
 - Add Unit test's for `Offer` and `Agreements` events


test/services/storage/storage.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+import { OfferService, StorageServices } from '../../../src/services/storage'
+import { sequelizeFactory } from '../../../src/sequelize'
+import Offer from '../../../src/services/storage/models/offer.model'
+
+chai.use(sinonChai)
+chai.use(chaiAsPromised)
+chai.use(dirtyChai)
+const expect = chai.expect
+
+const mockEventsFactory = (event: Record<string, unknown> = {}): EventData => Object.assign({
```
AuHau, line 21: There is pretty much same function here: https://github.com/rsksmart/rif-marketplace-cache/blob/master/test/utils.ts#L45

So please use this one so we don't replicate this unnecessarily.

test/services/storage/storage.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+        offerService = new OfferService({ Model: Offer })
+        offerServiceEmitSpy = sinon.spy()
+        offerService.emit = offerServiceEmitSpy
+        processor = eventProcessor({ offerService } as StorageServices, eth)
+      })
+      beforeEach(async () => {
+        await sequelize.sync({ force: true })
+      })
+
+      it('Should create new Offer if not existed', async () => {
```
AuHau, line 57: Please use lowercase "S". It sort of joins into a sentence with the `describe()` part: 
> Offer events should create...

test/services/storage/storage.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+          returnValues: {
+            provider: 'TestAddress',
+            totalCapacity: 1000
+          }
+        })
+        const eventFromDb = await Offer.create({ address: event.returnValues.provider })
+        await processor(event)
+        const updatedEventFromDB = await Offer.findOne({ where: { address: event.returnValues.provider } })
+
+        expect(updatedEventFromDB?.updatedAt).to.be.gt(eventFromDb.updatedAt)
```
AuHau, line 77: Hmm I would argue to test rather the exact values that changed (for example that the `totalCapacity` changed) then use `updatedAt`. Even though it is part of the model lets not depend on it.

src/services/storage/handlers/offer.ts
```diff
@@ -66,6 +66,8 @@ const handler: Handler<StorageServices> = {
 
     await offer.save()
 
+    // @TODO if we receive an unknown message then update actually doesn't happens, because offer don't changed
```
AuHau, line 69: Good catch, although I think it is not such a big deal if the `updated` event is emitted. Especially since this should not happen often. I guess leaving the TODO there, for now, is sufficient.

nduchak: Ok

test/services/storage/storage.spec.ts
nduchak: Sure, miss this one=)

src/services/storage/handlers/offer.ts
```diff
@@ -46,26 +47,28 @@ const handler: Handler<StorageServices> = {
...
         }
 
-        const [firstMsg, ...restMsg] = msg[0].replace('0x', '')
-        const flag = firstMsg.substring(0, 2)
+        const [firstMsg, ...restMsg] = msg
+        const flag = firstMsg.substring(2, 4)
 
         if (flag === '01') { // PeerId definition
-          offer.peerId = decodeByteArray([firstMsg.substring(2), ...restMsg])
+          offer.peerId = decodeByteArray([`0x${firstMsg.substring(4)}`, ...restMsg])
```
AuHau, line 54: Hmm you don't need to reconstruct the `0x` of the message. See how `hexToAscii` is implemented [here](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-utils/src/index.js#L168).
But since you have it here, you can already keep it I guess :-)

test/services/storage/storage.spec.ts
```diff
@@ -0,0 +1,202 @@
+import chai from 'chai'
```
AuHau, line 1: I would rename the file to `processor.spec.ts` as you are unit-testing the processor component here and it is already nested in `storage` folder ;-)

test/services/storage/storage.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+      let offerService: OfferService
+      let offerServiceEmitSpy: sinon.SinonSpy
+
+      before(async () => {
+        offerService = new OfferService({ Model: Offer })
+        processor = eventProcessor({ offerService } as StorageServices, eth)
+        await sequelize.sync({ force: true })
+      })
+      beforeEach(() => {
+        offerServiceEmitSpy = sinon.spy()
```
AuHau, line 46: I know I have this pattern as well in my tests, so just FYI if you want to be better then me 😅   then there is the [`spy.resetHistory()`](https://sinonjs.org/releases/v9.0.2/spies/#spyresethistory) call so you don't have to recreate and reassign the spy every test :-)

test/services/storage/storage.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+  describe('Events Processor', () => {
+    describe('Offer events', () => {
+      let processor: (event: EventData) => Promise<void>
+      let offerService: OfferService
+      let offerServiceEmitSpy: sinon.SinonSpy
+
+      before(async () => {
+        offerService = new OfferService({ Model: Offer })
+        processor = eventProcessor({ offerService } as StorageServices, eth)
+        await sequelize.sync({ force: true })
```
AuHau, line 43: Hmm, this means that you are depending on the state from previous tests right? Eq. not starting from "scratch" every time?

I am strongly against this as this effectively requires running the whole test-suite every time which is something that is not guaranteed and actually even desired. If you have test-suite running for 10 minutes and you are trying debugging one test, you really don't want to run it all, but actually only that one test, which is something you can actually do using the `--grep` functionality of Mocha. E.g. `npm test -- -g "update BillingPlan"' which will fail even though the tested code might be correct.

Also, it is hard for understanding the tests because you have to read a lot of other tests to understand what state are you testing.

And also, if one test fails then it will break all dependant tests which might be actually pass otherwise.

So please rework the tests in the way that every test-case runs from scratch, setups the state (eq. DB etc.) before the running the tested code and then assert that the state changed as expected.

I know it is tedious and very verbose but it pays out to have it like this, at least for unit-tests, integration tests might be a different story.

src/services/storage/handlers/offer.ts
```diff
@@ -46,26 +47,28 @@ const handler: Handler<StorageServices> = {
...
-          logger.error(`Unknown message flag ${flag}!`)
+          throw new EventError(`Unknown message flag ${flag}!`, event.event)
         }
         break
       }
       case 'BillingPlanSet':
         await updatePrices(offer, event.returnValues.period, event.returnValues.price)
         break
       default:
-        logger.error(`Unknown event ${event.event}`)
```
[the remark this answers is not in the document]
nduchak: So basically this is a dead code as we have an event filtering [here](https://github.com/rsksmart/rif-marketplace-cache/blob/master/src/services/storage/storage.processor.ts#L14) and all unknown events already ignored.
Should I add some handler for unknown events?

test/services/storage/storage.spec.ts
nduchak: Good point, thanks 👍

src/services/storage/handlers/offer.ts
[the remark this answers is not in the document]
AuHau: Hmmm I see, you are right. Yet even though in runtime this might not happen then based on input data from the blockchain, for example during development there might be some inconsistency in configuration and lead to this? I admit that it is not so strong case, but even that I would still keep it there 😅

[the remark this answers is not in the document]
nduchak: Yes, don't think about configuration. It's reverted 👍

test/utils.ts
```diff
@@ -58,8 +58,15 @@ export function eventMock (options?: Partial<EventData>): EventData {
   return testEvent
 }
 
-export function blockMock (blockNumber: number, blockHash = '0x123'): BlockTransactionString {
+export function blockMock (blockNumber: number, blockHash = '0x123', options?: Partial<BlockTransactionString>): BlockTransactionString {
   const block = Substitute.for<BlockTransactionString>()
+  options = options || {}
```
vojtechsimetka, line 63:
```suggestion
export function blockMock (blockNumber: number, blockHash = '0x123', options: Partial<BlockTransactionString> = {}): BlockTransactionString {
  const block = Substitute.for<BlockTransactionString>()
```

test/utils.ts
```diff
@@ -58,8 +58,15 @@ export function eventMock (options?: Partial<EventData>): EventData {
...
+
+  for (const [key, value] of Object.entries(options)) {
+    // @ts-ignore
+    block[key].returns!(value)
+  }
+
```
vojtechsimetka, line 69:
```suggestion

Object.entries(options).forEach([key, value] => {
    // @ts-ignore
    block[key].returns!(value)
  })

```

nduchak: it's just copy-paste. I also prefer to use built-in Array methods)

PR #193, opened 2020-07-02
fix(storage): correct billing period unit calculation

AuHau:
Solidity timestamps = seconds
JS timestamps = milliseconds 

This PR fixes mismatch for billing period is in seconds and everything else in miliseconds.


AuHau:
Solved by #191

PR #216, opened 2020-07-20
feat: wrap and expose agreement and offer events

nduchak:


src/services/storage/handlers/offer.ts
```diff
@@ -68,9 +76,9 @@ const handler: Handler<StorageServices> = {
 
     await offer.save()
 
-    // @TODO if we receive an unknown message then update actually doesn't happens, because offer don't changed
-    // Should we emit `update` event then???
-    if (offerService.emit) offerService.emit(created ? 'created' : 'updated', offer.toJSON())
+    if (offerService.emit && created) {
```
AuHau, line 79: You have to emit `updated` event for the `BillingPlanSet` as well no?

test/services/storage/processor.spec.ts
```diff
@@ -290,7 +290,7 @@ describe('Storage services: Events Processor', () => {
         expect(agreement?.billingPrice).to.be.eql(plan.amount)
         expect(agreement?.availableFunds).to.be.eql(event.returnValues.availableFunds)
         expect(agreement?.lastPayout).to.be.eql(await getBlockDate(eth, event.blockNumber))
-        expect(agreementServiceEmitSpy).to.have.been.calledOnceWith('created')
+        expect(agreementServiceEmitSpy).to.have.been.calledWith('created')
```
AuHau, line 293: Interesting why did you change from `calledOnceWith` to `calledWith`? Shouldn't it be called only once?

src/services/storage/handlers/offer.ts
nduchak: Hmm, that's interesting. We are not updating offer in this case, should we send offer update event?
For me, it's not clear, if I receive offer update event but offer actually doesn't changed

AuHau: I see. Well it is true that we have Billing Plans as separate models. But they are presented as nested objects in the end point... 🤔 I guess toJSON() does not include associations.

It is true that we don't really need this information for the Pining Service, but maybe this should be included for the Marketplace UI?

nduchak: Yes, this make sense, and `toJSON ` includes associations. Then will emit it also for BillingPlans
Thanks 👍

src/services/storage/handlers/offer.ts
```diff
@@ -35,42 +75,17 @@ const handler: Handler<StorageServices> = {
...
 
+    await handlers[event.event](event, offer, offerService)
+
     await offer.save()
 
-    // @TODO if we receive an unknown message then update actually doesn't happens, because offer don't changed
-    // Should we emit `update` event then???
-    if (offerService.emit) offerService.emit(created ? 'created' : 'updated', offer.toJSON())
+    if (offerService.emit && created) {
+      offerService.emit('created', offer.toJSON())
```
AuHau, line 87: Well with this refactor, I believe we will get both `created` and `updated` events for the first event of the Offer right? Maybe we can just simply ditch the `created` for now and see if    the need for it raises later on?
I will leave that up to you to decide :-)

nduchak: hmm... you are right.

PR #221, opened 2020-07-26
feat: expose new-block event

nduchak:
Add NewBlockEmitter service and expose it on `/new-block` url

### DIFF
--- a/src/blockchain/index.ts
+++ b/src/blockchain/index.ts
@@ -1,11 +1,13 @@
 import config from 'config'
 import Eth, { BlockHeader } from 'web3-eth'
+import { getObject } from 'sequelize-store'
 
 import { ConfirmatorService } from './confirmator'
 import { Application, ServiceAddresses } from '../definitions'
 import { loggingFactory } from '../logger'
-import { NewBlockEmitterService } from './new-block-emitters'
+import { NEW_BLOCK_EVENT_NAME, NewBlockEmitterService } from './new-block-emitters'
 import { getNewBlockEmitter } from './utils'
+import { waitForReadyApp } from '../utils'
 
 const logger = loggingFactory('blockchain')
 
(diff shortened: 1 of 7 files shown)


src/blockchain/new-block-emitters.ts
```diff
@@ -127,3 +130,20 @@ export class ListeningNewBlockEmitter extends AutoStartStopEventEmitter {
     this.subscription?.unsubscribe(error => { this.logger.error(error) })
   }
 }
+
+export class NewBlockEmitterService implements Partial<ServiceMethods<any>> {
```
AuHau, line 134: Hmm can you call this service with GET request? To get the latest block number? I think there should be `get()` method where you return the `getLastBlockNumber` no?

nduchak: Hmm, as I understand `get`  method require `id` argument, `find` method should be just GET request.

AuHau: Yeah you are right. But still there is no find here right? 😅

nduchak: yes=). So which method should be the best to fetch last block?

AuHau: Well `find` I would say :-) Simple `GET` and it should return you something like `{ latestBlockNumber: ... }` I guess?

nduchak: yea, but if you use feather client on another side should you have a small confusion?

src/blockchain/index.ts
```diff
@@ -14,23 +18,44 @@ export function ethFactory (): Eth {
...
+  const eth = app.get('eth')
+  const store = getObject()
+  const newBlockEmitter = getNewBlockEmitter(eth)
+  const newBlockEmitterService = app.service(ServiceAddresses.NEW_BLOCK_EMITTER)
+
+  // Subscribe for new blocks
+  newBlockEmitter.on(NEW_BLOCK_EVENT_NAME, (block: BlockHeader) => {
+    logger.debug('New block: ', block)
+    newBlockEmitterService?.emit(NEW_BLOCK_EVENT_NAME, block)
+    store['blockchain.lastFetchedBlockNumber'] = block.number
```
AuHau, line 47: Oh I see it now. Why do you emit the whole block header but store only its number? I believe that you should store the whole object so you can `GET` the whole object instead of the only number.

nduchak: Done

src/blockchain/new-block-emitters.ts
```diff
@@ -127,3 +130,21 @@ export class ListeningNewBlockEmitter extends AutoStartStopEventEmitter {
...
+  constructor () {
+    this.events = [NEW_BLOCK_EVENT_NAME]
+  }
+
+  // eslint-disable-next-line require-await
+  async find (): Promise<BlockHeader> {
+    return this.getLastBlockNumber()
+  }
+
+  getLastBlockNumber (): BlockHeader {
```
AuHau, line 147: Small nitpick 😇 It is not "last block **number**" anymore. Can you please rename it?

nduchak: Oops

PR #222, opened 2020-07-27
feat: add new RNS contract

julianmrodri:
* Bump version of `RNS` contract to `v0.1.2`
* Use new `RNS` simple placement contract - Update Addresses.
* Ignore `Transfers` handler when coming form the Marketplace Contract 
* Process `transfers` related to the Marketplace contract in the `TokenSold` handler.
* Create `registerTransfer` function to avoid code duplication as Transfers are now generated in both handlers.

Works together with UI PR: https://github.com/rsksmart/rif-marketplace-ui/pull/271
Setup and Deploy locally - Dev PR: https://github.com/rsksmart/rif-marketplace-dev/pull/66

### DIFF
--- a/src/services/rns/rns.processor.ts
+++ b/src/services/rns/rns.processor.ts
@@ -3,7 +3,7 @@ import config from 'config'
 import { Eth } from 'web3-eth'
 import { EventData } from 'web3-eth-contract'
 import Utils from 'web3-utils'
-import { RnsServices } from '.'
+import { RnsService, RnsServices } from '.'
 import { getBlockDate } from '../../blockchain/utils'
 import { Logger } from '../../definitions'
 import DomainOffer from './models/domain-offer.model'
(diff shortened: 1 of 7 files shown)


src/services/rns/rns.processor.ts
```diff
@@ -13,12 +13,46 @@ import DomainOwner from './models/owner.model'
...
+  }
+}
+
 async function transferHandler (logger: Logger, eventData: EventData, eth: Eth, services: RnsServices): Promise<void> {
   const tokenId = Utils.numberToHex(eventData.returnValues.tokenId)
   const ownerAddress = eventData.returnValues.to.toLowerCase()
 
   const fiftsAddr = config.get('rns.fifsAddrRegistrar.contractAddress')
   const registrar = config.get('rns.registrar.contractAddress')
+  const marketplace = config.get('rns.placement.contractAddress')
```
AuHau, line 55:
```suggestion
  const marketplace = config.get<string>('rns.placement.contractAddress')
```

src/services/rns/rns.processor.ts
```diff
@@ -56,35 +90,12 @@ async function transferHandler (logger: Logger, eventData: EventData, eth: Eth,
...
-  const transferDomain = await Transfer.create({
-    id: transactionHash,
-    tokenId,
-    sellerAddress: from,
-    buyerAddress: ownerAddress
-  })
-
-  if (transferDomain) {
-    logger.info(`Transfer event: Transfer ${tokenId} created`)
+  if (ownerAddress === (marketplace as string).toLowerCase() || from === (marketplace as string).toLowerCase()) {
```
AuHau, line 93:
```suggestion
  if (ownerAddress === marketplace.toLowerCase() || from === marketplace.toLowerCase()) {
```

src/services/rns/rns.processor.ts
```diff
@@ -13,12 +13,46 @@ import DomainOwner from './models/owner.model'
 import Transfer from './models/transfer.model'
 import SoldDomain from './models/sold-domain.model'
 
+// registerTransfer: creates the transfer record and updates DomainOwner
```
AuHau, line 16:
```suggestion
/**
 * Creates the transfer record and updates DomainOwner
 */
```

src/services/rns/rns.processor.ts
```diff
@@ -3,7 +3,7 @@ import config from 'config'
 import { Eth } from 'web3-eth'
 import { EventData } from 'web3-eth-contract'
 import Utils from 'web3-utils'
-import { RnsServices } from '.'
+import { RnsService, RnsServices } from '.'
```
vojtechsimetka, line 6: Double import :)

src/services/rns/rns.processor.ts
julianmrodri: Nice thanks will include this change

src/services/rns/rns.processor.ts
julianmrodri: no.. one is in plural the other in singular. Because now im passing an individual service as parameter and need the type

julianmrodri:
Adam, it is backward compatible. But anyway we DO NOT need to be backward compatible as this will start from scratch as an new contract on the Testnet, and then on Mainnet. Future versions Might need to be Compatible for sure.

src/services/rns/rns.processor.ts
AuHau: Yeah, it is bit a bad naming. Maybe we should rename it to `RnsBaseService`? As the `RnsServices` are derived from that.

julianmrodri: Yes will do that.

julianmrodri:
Sure will fix the commit message and the rename.

src/services/rns/rns.processor.ts
julianmrodri: Renamed it.

PR #230, opened 2020-08-03
feat: new id column for multitransaction

### LINKED ISSUE #223: Handle multiple RNS Operations in same transaction
* Issue: The DB Model was design to support only a Single transaction model, which works fine if users ONLY operate through the UI tools we propose. But it is possible that users may want to interact directly with the Contract, placing or buying multiple domains on the same Transaction, which would cause our Cache to fail. 
* Solution: We need to fix this ASAP, the DB model should work independent of the Tx #. The Transaction Hash should not be the PK in the tables, but simply an additional attribute.

julianmrodri:
* include new `id` and `txHash` column autonumeric to handle multiple operations in single transaction and not depend on txID. This impacts offer, sold domains, and transfers.

Closes #223

### DIFF
--- a/src/services/rns/rns.processor.ts
+++ b/src/services/rns/rns.processor.ts
@@ -16,16 +16,16 @@ import SoldDomain from './models/sold-domain.model'
 /**
  * Creates the transfer record and updates DomainOwner
  */
-async function registerTransfer (transactionHash: string, tokenId: string, sellerAddress: string, buyerAddress: string, logger: Logger, domainsService: RnsBaseService) {
+async function registerTransfer (transactionHash: string, tokenId: string, sellerAddress: string, buyerAddress: string, logger: Logger, domainsService: RnsBaseService): Promise<number> {
   const transferDomain = await Transfer.create({
-    id: transactionHash,
+    txHash: transactionHash,
     tokenId,
     sellerAddress,
     buyerAddress
   })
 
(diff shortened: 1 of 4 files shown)


src/services/rns/rns.processor.ts
```diff
@@ -16,16 +16,16 @@ import SoldDomain from './models/sold-domain.model'
 /**
  * Creates the transfer record and updates DomainOwner
  */
-async function registerTransfer (transactionHash: string, tokenId: string, sellerAddress: string, buyerAddress: string, logger: Logger, domainsService: RnsBaseService) {
+async function registerTransfer (transactionHash: string, tokenId: string, sellerAddress: string, buyerAddress: string, logger: Logger, domainsService: RnsBaseService): Promise<number> {
```
jurajpiar, line 19: maybe just call it `txHash` so that the name doesn't change across the function.

jurajpiar: also the line is too long (see code smells).

julianmrodri: We can do this.

julianmrodri:
Well If all the current functionality works then we are already in better shape because the table does not rely on txId.  Thats already enough. But I take what yous ay and will try to include some tests like Storage is doing mocking the events. E2E tests definitly would leave it for future changes.

PR #243, opened 2020-08-04
feat(core): support listening on topics

AuHau:
Adds support for listening topics, which will filter only those events from contract that we are interested in. If the Contract emits other events that we are not interested then they will be ignored directly by the blockchain node and won't be even fetched to Cache.


AuHau:
True, maybe since they are optional anyway, I can move `topics` and `events` to `options` 🤔

PR #254, opened 2020-08-12
feat: bignumber support

nduchak:


src/services/storage/models/agreement.model.ts
```diff
@@ -41,30 +45,39 @@ export default class Agreement extends Model {
...
+      : bn(0)
   }
 
   @Column(DataType.VIRTUAL)
-  get periodsSinceLastPayout () {
-    return Math.floor((Date.now() - this.lastPayout.getTime()) / (this.billingPeriod * 1000))
+  get periodsSinceLastPayout (): BigNumber {
+    // Date.now = ms
+    // this.lastPayout.getTime = ms
+    // this.billingPeriod = seconds ==> * 1000
```
vojtechsimetka, line 63: Nice!

src/utils.ts
```diff
@@ -4,11 +4,23 @@ import { readFile as readFileCb } from 'fs'
...
+import BigNumber from 'bignumber.js'
 
 import { Application, Config, isSupportedServices, Logger, SupportedServices } from './definitions'
 
 const readFile = promisify(readFileCb)
 
+/**
+ * Bignumber.js utils functions
+ */
+export function bn (v: any): BigNumber {
```
vojtechsimetka, line 16: Please restrict the type of the argument, to `string`, `number` or `BigNumber`

src/utils.ts
```diff
@@ -4,11 +4,23 @@ import { readFile as readFileCb } from 'fs'
...
+  return new BigNumber(v)
+}
+
+export function bnFloor (v: any): BigNumber {
```
vojtechsimetka, line 20: Please restrict the type of the argument, to `string`, `number` or `BigNumber`

src/utils.ts
vojtechsimetka, line 16: I am not a big fan of this utility function but the call is on @AuHau & @jurajpiar . To me it hinders readability and you can use `new BigNumber(...)` or even just `BigNumber(...)`

nduchak: Fixed

nduchak: fixed

PR #256, opened 2020-08-13
feat: apply storage contract changes

### LINKED ISSUE #253: Implement the contract changes in cache

### LINKED ISSUE #255: Make PeerId required for Offers
As there is plan to integrate communication between the Pinning Service and the Consumer for error and progress reporting should we make the PeerId for the Provider required.

As creation of Offer consists of emitting multiple events that are mandatory (`TotalCapacitySet` and `BillingPlanSet`) we should add also the PeerId message also to the mandatory events.

nduchak:
filter out offers without `peerid`
Closes #255 and closes #253

### DIFF
--- a/config/default.json5
+++ b/config/default.json5
@@ -10,6 +10,17 @@
   },
 
   blockchain: {
+    // Topics that will be listened to
+    topics: [
+      'TotalCapacitySet(address,uin64)',
+      //      'BillingPlanSet(address,uint64,uint128)',
+      'MessageEmitted(address,bytes32[])',
+      'NewAgreement(bytes32,bytes32[],address,address,uint128,uint64,uint128,uint256)',
+      'AgreementFundsDeposited(bytes32,uint256)',
+      'AgreementFundsWithdrawn(bytes32,uint256)',
+      'AgreementFundsPayout(bytes32,uint256)',
+      'AgreementStopped(bytes32)',
+    ],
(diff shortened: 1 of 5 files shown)


config/default.json5
```diff
@@ -10,6 +10,17 @@
   },
 
   blockchain: {
+    // Topics that will be listened to
+    topics: [
```
nduchak, line 14: Do we actually need that here?

src/services/storage/models/agreement.model.ts
```diff
@@ -20,6 +20,7 @@ export default class Agreement extends Model {
   @Column({ type: DataType.STRING(64) })
   consumer!: string
 
+  // In Megabytes
```
AuHau, line 23: Interesting, I guess since we don't do any calculations with external data we don't have to change it anywhere else right? 🤔

nduchak: You are right. Just add comments for future

config/default.json5
nduchak: As I see, cache don't have the support of topics filtering, as we had in pinner

AuHau: lol you are right, the porting PR is still to be merged :-D I will merge it then :-)

nduchak: =)

Issue #1, opened 2020-01-30
DB Migrations


AuHau:
Before release, there should be created proper migrations:

https://github.com/sequelize/cli/blob/master/docs/README.md
https://sequelize.org/v5/manual/migrations.html
!!! https://github.com/feathersjs-ecosystem/feathers-sequelize#migrations

AuHau:
Heh it is already "critical" 😅 Can't get much higher than this 😝
I will work on it hopefully on beginning of next week.

julianmrodri:
Yes no worries. I thought it was mandatory for Testnet but we can work around it.

julianmrodri:
Actually shoulnt be Critical

julianmrodri:
I think we can close this for some reason didnt ge tclose when merging PR.

AuHau:
Yeah, we can. It was only mentioned in the PR not [linked](https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references)

PR #269, opened 2020-08-18
feat: db migrations

nduchak:
add db-migration command to cli


nduchak:
Closes #1

src/cli/db-migration.ts
```diff
@@ -0,0 +1,154 @@
+import fs from 'fs'
```
AuHau, line 1: This command should replace the `db-sync` command. Not sure if this command should be renamed to `db-sync` for backward compatibility or just remove the `db-sync` and keep  `db-migrations` 🤔 I will leave that to your decision.

src/cli/db-migration.ts
```diff
@@ -0,0 +1,154 @@
...
+      exclusive: ['down', 'generate']
+    }),
+    down: flags.boolean({
+      char: 'd',
+      description: 'Undo db migrations',
+      exclusive: ['up', 'generate']
+    }),
+    generate: flags.string({
+      char: 'd',
+      description: 'Generate migrations',
```
AuHau, line 43: This description is not helpful at all. It requires some string, but I don't know what I should put there? I thought that it is just a boolean flag originally.

AuHau: I also thought that this generate actual migration and not just a template. Maybe this info should be given as well.

migrations/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    return this.umzugIns.on
+  }
+
+  // eslint-disable-next-line require-await
+  async pending (): Promise<Umzug.Migration[]> {
+    return this.umzugIns.pending().catch(e => {
+      if (e.code === 'ENOENT') {
+        return []
+      }
+      throw e
```
AuHau, line 48: This is not in `async` function, so you should not use `throw` here.

package.json
```diff
@@ -64,6 +64,7 @@
     "@rsksmart/rns-auction-registrar": "1.0.2",
     "@rsksmart/rns-reverse": "^1.0.2",
     "@rsksmart/rns-rskregistrar": "^1.2.1",
+    "@types/umzug": "^2.2.3",
```
AuHau, line 67: This should be in `devDependencies`

src/sequelize.ts
```diff
@@ -51,12 +52,16 @@ export function BigNumberStringType (propName: string): Partial<ModelAttributeCo
...
-export default function (app: Application): void {
+export default async function (app: Application): Promise<void> {
   const sequelize = sequelizeFactory()
   const oldSetup = app.setup
 
   app.set('sequelize', sequelize)
 
+  // Run migration
+  logger.info('Run DB Migrations')
+  await DbMigration.getInstance(sequelize).up()
```
AuHau, line 63: You should not run migrations automatically! 
There should be a check here if there are no pending migrations and if so then throw an error, exit and let the user handle this situation with the `db-migration` command!

AuHau (changes requested):
Hmm there are some changes needed here that I missed in the previous PR in Pinner. We should backport those changes to Pinner as well!

Also, I believe that this PR should contain the first migration that will bootstrap the DB into the current state!

src/cli/db-migration.ts
nduchak: Ok. Probably I remove db-sync, and leave migration command as it is

src/cli/db-migration.ts
nduchak: Done

migrations/index.ts
nduchak: Done

package.json
nduchak: Done

src/sequelize.ts
```diff
@@ -51,12 +52,16 @@ export function BigNumberStringType (propName: string): Partial<ModelAttributeCo
   }
 }
 
-export default function (app: Application): void {
+export default async function (app: Application): Promise<void> {
```
AuHau, line 55: Interesting that linter doesn't yell about not using `await` 🤔

nduchak: Had `await` here: https://github.com/rsksmart/rif-marketplace-cache/pull/269/files/a34fa72a7815c31bd81e6a900e46e2f4dcf090f6#diff-f39990cee8f2282dd02df78076ac6a13R61

AuHau: Oh missed that one 😅

AuHau (approved):
LGTM! But maybe I would wait with merge until the `release/mainnet` is done and then regenerated the initial migration and start versioning DB after that 🤔

AuHau (approved):
LGTM!

PR #270, opened 2020-08-20
fix: release-mainnet - update npm dependencies

julianmrodri:
* Update npm dependencies for the new `release/mainnet` branch.


julianmrodri:
@AuHau  any idea why it would fail for `node 10` ?

PR #271, opened 2020-08-21
Feat/rns enhancements

### LINKED ISSUE #223: Handle multiple RNS Operations in same transaction

julianmrodri:
* Updates RNS contract version to `0.1.4`
* Handles multiple operations in same transaction
* Add handler for `Approval` event to prevent UI Data pollution (adding the `approved` field to the Offers)
* Includes unit tests for the complete `RNS` module.

Closes #223


config/test.json5
```diff
@@ -3,5 +3,23 @@
   blockchain: {
     provider: 'ws://localhost:8545',
     waitBlockCountBeforeConfirmationRemoved: 10
+  },
+  rns: {
+    enabled: true,
+    owner: {
+      contractAddress: "OWNER_ADDR"
```
AuHau, line 10: What is the reason for these placeholders? Failing tests?

src/services/rns/models/domain-offer.model.ts
```diff
@@ -4,8 +4,11 @@ import Domain from './domain.model'
 
 @Table({ freezeTableName: true, tableName: 'rns_domain-offer', timestamps: false })
 export default class DomainOffer extends Model {
-  @Column({ primaryKey: true, type: DataType.STRING })
-  offerId!: string
+  @Column({ primaryKey: true, autoIncrement: true, type: DataType.INTEGER })
```
AuHau, line 7: Just FYI thanks to #269 we will have to start writing DB migrations, so before the `release/mainnet` will be merged to `master` it will have to include migrations for these changes as well.

AuHau (commented):
I will leave the approval to @jurajpiar so just few comments here.

Btw. if I understand it correctly you won't show Domain until it is `approved` right? Maybe you can utilized same mechanism as there is in Storage, where until it is "valid" (eq. some conditions are met) the Offer is not listed in the results. It is achieved using DB Scopes see: 

 - here is scope definition: https://github.com/rsksmart/rif-marketplace-cache/blob/master/src/services/storage/models/offer.model.ts#L9
 - here is the scope usage: https://github.com/rsksmart/rif-marketplace-cache/blob/master/src/services/storage/hooks/offers.hooks.ts#L21

The Scope usage part can be even skipped if you would use `@DefaultScope()` on the model definition.

config/test.json5
julianmrodri: Right, we needed the addresses defined in the `test.json5` file so they can be picked in the Unit tests, because they are actually used in the Even HAndlers.

src/services/rns/models/domain-offer.model.ts
julianmrodri: Even if it is the first deployment? Is this something we can easily integrate? @nduchak

AuHau: Well the merging of #269 defines the point from where we have to start writing migrations, because it includes the first migration that bootstrap the database into the shape which is the point of the merge. And since this will be most probably be merged after that than yes, this should have migraitons.

julianmrodri: Ok Ok.. I got confused.. you mentioned that as a requirement to merge to `master` but not to DEPLOY to Mainnet right?

AuHau: Yup as merge to `master` :-)

I mean it would be awesome to have it already for deployment for `mainnet` but I can understand that it won't be added. With these migrations, the DevOps will be able to do incremental deployments where the whole "remove DB and precache" phase won't be needed anymore...

config/test.json5
julianmrodri: Is this OK? Do you see a better way to mockup these config file values for the Tests?

src/services/rns/models/domain-offer.model.ts
julianmrodri: Understood.

julianmrodri:
Resubmitted using Scopes and fixing some refactors. We decided to tackle code redundancy in the Tests in later PR.

jurajpiar (approved):
LG

PR #286, opened 2020-09-02
feat(storage): staking support

nduchak:
This PR introduces listening and handling of Staking Contract Events

### DIFF
--- a/src/cli/start.ts
+++ b/src/cli/start.ts
@@ -3,7 +3,7 @@ import { flags } from '@oclif/command'
 
 import { appFactory, services } from '../app'
 import { loggingFactory } from '../logger'
-import { Flags, Config, SupportedServices, isSupportedServices } from '../definitions'
+import { Flags, Config, SupportedServices, isSupportedServices, Application } from '../definitions'
 import { BaseCLICommand } from '../utils'
 import DbBackUpJob from '../db-backup'
 import { ethFactory } from '../blockchain'
@@ -108,13 +108,13 @@ ${formattedServices}`
       const resetPromise = new Promise(resolve => {
         appFactory({
           appResetCallBack: () => resolve()
-        }).then(({ app, stop }) => {
+        }).then((application: { app: Application, stop: () => void }) => {
           // Lets save the function that stops the app
-          stopCallback = stop
+          stopCallback = application.stop
 
           // Start server
           const port = config.get('port')
-          const server = app.listen(port)
+          const server = application.app.listen(port)
 
           server.on('listening', () =>
             logger.info(`Server started on port ${port}`)
--- a/src/cli/start.ts
+++ b/src/cli/start.ts
@@ -108,13 +108,13 @@ ${formattedServices}`
       const resetPromise = new Promise(resolve => {
         appFactory({
           appResetCallBack: () => resolve()
-        }).then((application: { app: Application, stop: () => void }) => {
+        }).then(({ app, stop }) => {
           // Lets save the function that stops the app
-          stopCallback = application.stop
+          stopCallback = stop
 
           // Start server
           const port = config.get('port')
-          const server = application.app.listen(port)
+          const server = app.listen(port)
 
           server.on('listening', () =>
             logger.info(`Server started on port ${port}`)
--- a/src/cli/start.ts
+++ b/src/cli/start.ts
@@ -3,7 +3,7 @@ import { flags } from '@oclif/command'
 
 import { appFactory, services } from '../app'
 import { loggingFactory } from '../logger'
-import { Flags, Config, SupportedServices, isSupportedServices, Application } from '../definitions'
+import { Flags, Config, SupportedServices, isSupportedServices } from '../definitions'
 import { BaseCLICommand } from '../utils'
 import DbBackUpJob from '../db-backup'
 import { ethFactory } from '../blockchain'
--- a/src/services/storage/handlers/stake.ts
+++ b/src/services/storage/handlers/stake.ts
@@ -14,7 +14,7 @@ const logger = loggingFactory('storage:handler:stake')
  * Supported tokens
   */
 type SupportedTokens = 'rbtc' | 'rif' | 'cfp'
-export const SUPPORTED_TOKENS_SYMBOLS = ['rbtc', 'rif', 'cfp']
+
 /**
  * TODO replace with some import
  */
--- a/src/services/storage/handlers/stake.ts
+++ b/src/services/storage/handlers/stake.ts
@@ -13,12 +13,16 @@ const handlers = {
 
     const stake = await StakeModel.findOne({ where: { token, account } })
 
-    stake!.total = total
-    await stake!.save()
+    if (!stake) {
+      throw new Error(`Stake for acoount ${account}, token ${token} not created`)
+    }
+
+    stake.total = total
+    await stake.save()
     logger.info(`Account ${account} stake amount ${amount}, final balance ${total}`)
 
     if (stakeService.emit) {
-      stakeService.emit('updated', stake!.toJSON())
+      stakeService.emit('updated', stake.toJSON())
     }
   },
 
@@ -27,8 +31,12 @@ const handlers = {
 
     const stake = await StakeModel.findOne({ where: { token, account } })
 
-    stake!.total = total
-    await stake!.save()
+    if (!stake) {
+      throw new Error(`Stake for acoount ${account}, token ${token} not created`)
+    }
+
+    stake.total = total
+    await stake.save()
     logger.info(`Account ${account} stake amount ${amount}, final balance ${total}`)
 
     if (stakeService.emit) {
--- a/src/services/storage/handlers/stake.ts
+++ b/src/services/storage/handlers/stake.ts
@@ -1,5 +1,4 @@
 import { EventData } from 'web3-eth-contract'
-import { Eth } from 'web3-eth'
 
 import { loggingFactory } from '../../../logger'
 import { Handler } from '../../../definitions'
@@ -10,21 +9,21 @@ const logger = loggingFactory('storage:handler:stake')
 
 const handlers = {
   async Staked (event: EventData, { stakeService }: StorageServices): Promise<void> {
-    const { user: offerId, total } = event.returnValues
+    const { user: account, total, data, amount } = event.returnValues
 
-    // TODO Update specific stake based on currency
-    const [, [stake]] = await StakeModel.update({ total }, { where: { offerId } })
-    logger.info(`Update stake for offer = ${offerId}`)
+    // TODO Update specific stake based on token
+    const [, [stake]] = await StakeModel.update({ total }, { where: { account, tokenAddress: data } })
+    logger.info(`Account ${account} stake amount ${amount}, final balance ${total}`)
 
     if (stakeService.emit) stakeService.emit('updated', stake.toJSON())
   },
 
-  async Unstacked (event: EventData, { stakeService }: StorageServices): Promise<void> {
-    const { user: offerId, total } = event.returnValues
+  async Unstaked (event: EventData, { stakeService }: StorageServices): Promise<void> {
+    const { user: account, total, data, amount } = event.returnValues
 
-    // TODO Update specific stake based on currency
-    const [, [stake]] = await StakeModel.update({ total }, { where: { offerId } })
-    logger.info(`Update stake for offer = ${offerId}`)
+    // TODO Update specific stake based on token
+    const [, [stake]] = await StakeModel.update({ total }, { where: { account, tokenAddress: data } })
+    logger.info(`Account ${account} unstack amount ${amount}, final balance ${total}`)
 
     if (stakeService.emit) stakeService.emit('updated', stake.toJSON())
   }
@@ -35,7 +34,7 @@ function isValidEvent (value: string): value is keyof typeof handlers {
 }
 
 const handler: Handler<StorageServices> = {
-  events: ['Stacked', 'Unstacked'],
+  events: ['Staked', 'Unstaked'],
   process (event: EventData, services: StorageServices): Promise<void> {
     if (!isValidEvent(event.event)) {
       return Promise.reject(new Error(`Unknown event ${event.event}`))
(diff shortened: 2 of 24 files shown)


AuHau (approved):

src/cli/start.ts
```diff
@@ -108,13 +108,13 @@ ${formattedServices}`
...
-        }).then(({ app, stop }) => {
+        }).then((application: { app: Application, stop: () => void }) => {
           // Lets save the function that stops the app
-          stopCallback = stop
+          stopCallback = application.stop
 
           // Start server
           const port = config.get('port')
-          const server = app.listen(port)
+          const server = application.app.listen(port)
```
AuHau, line 117: Why prepending with `application`? `application.app` does not look nice 😅

AuHau (changes requested):
LGTM, except on the handler structure, that I don't like.

src/services/storage/handlers/stake.ts
```diff
@@ -11,13 +12,13 @@ const handlers = {
   async Staked (event: EventData, { stakeService }: StorageServices): Promise<void> {
     const { user: account, total, token, amount } = event.returnValues
 
-    const stake = await StakeModel.findOne({ where: { token, account } })
+    let stake = await StakeModel.findOne({ where: { token, account } })
```
AuHau, line 15: Maybe a cleaner solution here would be [`findOrCreate`](https://sequelize.org/v5/class/lib/model.js~Model.html#static-method-findOrCreate)?

src/services/storage/handlers/stake.ts
```diff
@@ -29,18 +30,18 @@ const handlers = {
   async Unstaked (event: EventData, { stakeService }: StorageServices): Promise<void> {
     const { user: account, total, token, amount } = event.returnValues
 
-    const stake = await StakeModel.findOne({ where: { token, account } })
+    let stake = await StakeModel.findOne({ where: { token, account } })
 
     if (!stake) {
```
AuHau, line 35: Well, that does not really make sense, right? Can you get `Unstaked` event for `token` and `account` that did not call `Staked` before? I would keep the error here.

src/services/storage/index.ts
```diff
@@ -77,8 +78,8 @@ function precacheContract (eventEmitter: EventEmitter, services: StorageServices
 
 async function precache (possibleEth?: Eth): Promise<void> {
   const eth = possibleEth || ethFactory()
-  const storageEventsEmitter = getEventsEmitterForService(`${SERVICE_NAME}.${STORAGE_MANAGER}`, eth, storageManagerContract.abi as AbiItem[])
-  const stakingEventsEmitter = getEventsEmitterForService(`${SERVICE_NAME}.${STAKING}`, eth, stakingContract.abi as AbiItem[])
+  const storageEventsEmitter = getEventsEmitterForService(`${STORAGE_MANAGER}`, eth, storageManagerContract.abi as AbiItem[])
+  const stakingEventsEmitter = getEventsEmitterForService(`${STAKING}`, eth, stakingContract.abi as AbiItem[])
```
AuHau, line 82: No need for the `${}` part.
```suggestion
  const storageEventsEmitter = getEventsEmitterForService(STORAGE_MANAGER, eth, storageManagerContract.abi as AbiItem[])
  const stakingEventsEmitter = getEventsEmitterForService(STAKING, eth, stakingContract.abi as AbiItem[])
```

src/services/storage/index.ts
```diff
@@ -125,28 +126,28 @@ const storage: CachedService = {
     const reorgEmitterService = app.service(ServiceAddresses.REORG_EMITTER)
 
     // We require services to be precached before running server
-    if (!isServiceInitialized(`${SERVICE_NAME}.${STORAGE_MANAGER}`)) {
-      return logger.critical('Storage service is not initialized! Run precache command.')
+    if (!isServiceInitialized(`${STORAGE_MANAGER}`) || !isServiceInitialized(STAKING)) {
```
AuHau, line 129:
```suggestion
    if (!isServiceInitialized(STORAGE_MANAGER) || !isServiceInitialized(STAKING)) {
```

src/services/storage/index.ts
```diff
@@ -125,28 +126,28 @@ const storage: CachedService = {
...
+      return storageLogger.critical('Storage service is not initialized! Run precache command.')
     }
 
     const eth = app.get('eth') as Eth
     const confirmationService = app.service(ServiceAddresses.CONFIRMATIONS)
 
     // Storage Manager watcher
-    const storageManagerEventsEmitter = getEventsEmitterForService(`${SERVICE_NAME}.${STORAGE_MANAGER}`, eth, storageManagerContract.abi as AbiItem[])
-    storageManagerEventsEmitter.on('newEvent', errorHandler(eventProcessor(services, eth), logger))
+    const storageManagerEventsEmitter = getEventsEmitterForService(`${STORAGE_MANAGER}`, eth, storageManagerContract.abi as AbiItem[])
```
AuHau, line 137:
```suggestion
    const storageManagerEventsEmitter = getEventsEmitterForService(STORAGE_MANAGER, eth, storageManagerContract.abi as AbiItem[])
```

src/services/storage/index.ts
```diff
@@ -125,28 +126,28 @@ const storage: CachedService = {
...
+      storageManagerLogger.error(`There was unknown error in Events Emitter! ${e}`)
     })
     storageManagerEventsEmitter.on('newConfirmation', (data) => confirmationService.emit('newConfirmation', data))
     storageManagerEventsEmitter.on('invalidConfirmation', (data) => confirmationService.emit('invalidConfirmation', data))
     storageManagerEventsEmitter.on(REORG_OUT_OF_RANGE_EVENT_NAME, (blockNumber: number) => reorgEmitterService.emitReorg(blockNumber, 'storage'))
 
     // Staking watcher
-    const stakingEventsEmitter = getEventsEmitterForService(`${SERVICE_NAME}.${STAKING}`, eth, stakingContract.abi as AbiItem[])
-    stakingEventsEmitter.on('newEvent', errorHandler(eventProcessor(services, eth), logger))
+    const stakingEventsEmitter = getEventsEmitterForService(`${STAKING}`, eth, stakingContract.abi as AbiItem[])
```
AuHau, line 147:
```suggestion
    const stakingEventsEmitter = getEventsEmitterForService(STAKING, eth, stakingContract.abi as AbiItem[])
```

src/services/storage/index.ts
```diff
@@ -43,13 +43,14 @@ export interface StorageServices {
...
 
-const SERVICE_NAME = 'storage'
-const STORAGE_MANAGER = 'storageManager'
-const STAKING = 'staking'
+const STORAGE_MANAGER = 'storage.storageManager'
+const STAKING = 'storage.staking'
 
-const logger = loggingFactory(SERVICE_NAME)
+const storageLogger = loggingFactory('storage')
+const storageManagerLogger = loggingFactory(STORAGE_MANAGER)
```
AuHau, line 50: Great idea to different the loggers!

AuHau (changes requested):
Few suggestions

src/services/storage/handlers/stake.ts
nduchak: Good idea

src/services/storage/handlers/stake.ts
nduchak: Yea, theoretically it can not. But let's keep it there if you don't mind

AuHau: Hmmm, I don't like this. Because you are making invalid state a valid one, instead allowing us to detect it (as the error will be logged). Yes there is nothing you can do really about it (and in those cases the right thing to do is usually error-out), but we should find out about these problems and investigate.

Out of curiousity I have wrote some tests for this case in to the SC: https://github.com/rsksmart/rif-marketplace-storage/pull/127

And also with this approach you will be getting actually negative total amount...

So I would say, throw an error ;-)

AuHau:
Btw. this PR is blocked until the Staking PR is merged and released :-)

src/services/storage/handlers/stake.ts
nduchak: I am confused, as I am throwing an error :)

AuHau: Oh, sorry, you are rightfully confused 😅 Not sure why I thought that you are actually creating new entry there 😳
Sorry! My bad 🙈

AuHau (commented):
LGTM, now we only have to release the Staking part of the SC.

src/services/storage/handlers/stake.ts
nduchak: Probably that's because of force push which made before, so it's my bad=)

nduchak:
Blocked by: https://github.com/rsksmart/rif-marketplace-storage/pull/112

nduchak:
Closes #286 284

src/services/storage/handlers/stake.ts
```diff
@@ -5,18 +5,261 @@ import { loggingFactory } from '../../../logger'
...
 import StakeModel from '../models/stake.model'
+import { Eth } from 'web3-eth'
+import { AbiItem } from 'web3-utils'
 
 const logger = loggingFactory('storage:handler:stake')
 
+/**
+ * Supported tokens
+  */
+type SupportedTokens = 'rbtc' | 'rif' | 'cfp'
```
AuHau, line 16: Hmmmm, shouldn't this be specified in config?

src/services/storage/handlers/stake.ts
```diff
@@ -5,18 +5,261 @@ import { loggingFactory } from '../../../logger'
...
+
+/**
+ * Make a call to ERC20 token SC and return token symbol
+ * Return `rbtc` for ZERO_ADDRESS
+ * @param token
+ * @param eth
+ * @param tokenAbi
+ * @returns {SupportedTokens} token symbol
+ */
+function getTokenSymbol (token: string, eth: Eth, tokenAbi: AbiItem[] = ERC20Abi): Promise<SupportedTokens> {
```
AuHau, line 232: Can we make this function async please?

AuHau: And also `token` is not really token but `tokenContractAddress` here no?

src/services/storage/handlers/stake.ts
```diff
@@ -5,18 +5,261 @@ import { loggingFactory } from '../../../logger'
...
+  if (token === ZERO_ADDRESS) {
+    return Promise.resolve('rbtc')
+  }
+
+  const contract = new eth.Contract(tokenAbi, token)
+  return contract.methods.symbol().call({ from: eth.accounts.create() })
```
AuHau, line 238: Ufff, so you are calling the ERC20 contract here to resolve for given token address its symbol here? I am not sure this is a good approach. How fast is this call? Also, how do you make sure that this is actually value defined in `SupportedTokens`? 

I would say first of all we want have defined list of tokens that we support (most probably already mentioned `config` should be good place to have it there) and since we gonna do that we can add their address as well? We already do this for every other interaction with blockchain, so I would say it is standard and it will saves us lot of this work, that I am not sure if actually bring many benefits.

test/services/storage/models.spec.ts
```diff
@@ -201,6 +204,58 @@ const AGREEMENT_TEST_SCHEMA = [
...
 
 describe('Models', () => {
-  before(() => sequelizeFactory())
+  const sequelize = sequelizeFactory()
   describe('Agreement', () => generateModelGettersTests(AGREEMENT_TEST_SCHEMA, agreementFactory))
+  describe('Offer', () => {
+    beforeEach(async () => {
+      await sequelize.sync({ force: true })
+    })
+    it('Can aggregate total stakes for offers and order by stakes', async () => {
```
AuHau, line 213:
```suggestion
    it('should aggregate total stakes for offers and order by stakes', async () => {
```

src/services/storage/models/stake.model.ts
```diff
@@ -3,13 +3,15 @@ import BigNumber from 'bignumber.js'
 
 import { BigNumberStringType } from '../../../sequelize'
 
+export const SUPPORTED_TOKENS_SYMBOLS = ['rbtc', 'rif', 'cfp']
```
AuHau, line 6: This is definitely place to have it. Either `config` or `definitions.ts`, as already mentioned I prefer the config with the token's address.

src/services/storage/models/offer.model.ts
```diff
@@ -52,3 +54,21 @@ export default class Offer extends Model {
     return this.totalCapacity.minus(this.utilizedCapacity)
   }
 }
+
+export async function getStakesAggregateQuery (sequelize: Sequelize, currency: 'usd' | 'eur' | 'btc' = 'usd') {
```
AuHau, line 58: Can you add JSDoc explaining what this does?

test/utils.ts
```diff
@@ -82,6 +79,6 @@ export function rmDir (folder: string): void {
       fs.unlinkSync(path.join(folder, file))
     }
 
-    fs.rmdirSync(config.get<DbBackUpConfig>('dbBackUp').path, { recursive: true })
+    fs.rmdirSync(folder, { recursive: true })
```
AuHau, line 82: Good catch, although don't understand why you have the above `if` with `for`? Shouldn't the `{recursive: true}` handle this for you?

AuHau (changes requested):
I don't really agree with the approach for fetching the token's symbol. I would just use config and have configured tokens and their addresses. Simplicity FTW here...

src/services/storage/handlers/stake.ts
nduchak: Yes, will rename it

src/services/storage/handlers/stake.ts
nduchak: Hmm... The call is fast enough (it's static call), but I agree that tokens in the config would be better

src/services/storage/models/offer.model.ts
nduchak: Will do

test/utils.ts
nduchak: `{recursive: true}` can not remove folder if folder contains some files

AuHau (approved):
LGTM, thanks for implementing the suggestions!

PR #293, opened 2020-09-09
feat: batch support for domain registration

julianmrodri:
* Decode domain names when processed in Batch. This requires processing the internal transactions of the TX data field, using RLP and following the `FifsRegistrar` encoding.)
* Add support and configuration for the RNS Batch Registration contract (in all environments).
* Test for transaction decoding in Batch Mode (Registering two domains).

### DIFF
--- a/src/services/rns/rns.processor.ts
+++ b/src/services/rns/rns.processor.ts
@@ -12,6 +12,7 @@ import DomainExpiration from './models/expiration.model'
 import DomainOwner from './models/owner.model'
 import Transfer from './models/transfer.model'
 import SoldDomain from './models/sold-domain.model'
+import RLP = require('rlp')
 
 /**
  * Updates Domain Owner
@@ -67,10 +68,10 @@ async function registerTransfer (
 async function transferHandler (logger: Logger, eventData: EventData, eth: Eth, services: RnsServices): Promise<void> {
   const tokenId = Utils.numberToHex(eventData.returnValues.tokenId)
   const ownerAddress = eventData.returnValues.to.toLowerCase()
-
-  const fiftsAddr = config.get('rns.fifsAddrRegistrar.contractAddress')
-  const registrar = config.get('rns.registrar.contractAddress')
+  const fifsAddr = config.get<string>('rns.fifsAddrRegistrar.contractAddress')
+  const registrar = config.get<string>('rns.registrar.contractAddress')
   const marketplace = config.get<string>('rns.placement.contractAddress')
+  const batchprocess = config.get<string>('rns.batchContractAddress')
   const tld = config.get('rns.tld')
 
   const transactionHash = eventData.transactionHash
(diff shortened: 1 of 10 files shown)


jurajpiar (approved):
🚀 Looking good

src/services/rns/rns.processor.ts
```diff
@@ -64,13 +65,39 @@ async function registerTransfer (
...
+      const rlpDecoded: RLPDecoded = RLP.decode(decodedData.params[2].value) as unknown as RLPDecoded
+      const domainNames = rlpDecoded[1].map(
+        (domainData: number[]) =>
+          Utils.hexToAscii(
+            Utils.bytesToHex(
+              domainData.slice(88, domainData.length)
+            )
+          )
+      )
+      name = domainNames.find((domain: string) => Utils.numberToHex(Utils.sha3(domain) as string) === tokenId)
```
jurajpiar, line 84: this like could be cleaner if split on new lines

src/services/rns/rns.processor.ts
```diff
@@ -64,13 +65,39 @@ async function registerTransfer (
...
+    } else {
+      // Single registration
+      name = Utils.hexToAscii('0x' + decodedData.params[2].value.slice(218, decodedData.params[2].value.length))
```
jurajpiar, line 87: this like could be cleaner if separated to new lines

src/services/rns/rns.processor.ts
julianmrodri: Yes I did that. Thanks

src/services/rns/rns.processor.ts
julianmrodri: I did that in this new change. Looks much better assigning the `params` attribute to a shorter variable.

AuHau (approved):
I have taken the liberty to do small refactor for some style tweaks as I did not find any critical problems.

I have few general remarks though:
  - please don't force push once a PR is reviewed, it makes it hard to track the incremental changes for the reviewer
 - use `camelCase` for variable naming (`batchprocess` vs `batchProcess`)
 - if you want to write a comment for function use JSDoc (see [example here](https://github.com/rsksmart/rif-marketplace-cache/pull/293/commits/e267fd8dfcd685a60764989ad0bb99a577b36482#diff-6db85d3479d9df3cf870a47f5803e8cbL68)

One more thing that we should generally improve. Whenever there is some "magic" thing that cannot be deduced from the codebase itself, there should be an exhaustive comment on why it is what it is. For example when you have [`name = Utils.hexToAscii('0x' + domainData.slice(218, domainData.length))`](https://github.com/rsksmart/rif-marketplace-cache/pull/293/files#diff-6db85d3479d9df3cf870a47f5803e8cbR98) you should explain why do you `slice` from `218` (for example what is the format of the hash/data etc.). Similarly, why do you decode the Batch registration in the way you do.

The problem here is that if somebody who does not have a clue what is this about comes to the code, he won't have any idea what is going on. In my experience, this can be even you once you come back to the code in a month or two, after you moved to do some other thing, so it is generally good to have it there. We can add it in separate PR, so it does not block the release.

PR #327, opened 2020-10-06
feat: handle Billing plans remove

nduchak:


src/services/storage/handlers/offer.ts
```diff
@@ -75,7 +81,10 @@ const handlers: { [key: string]: Function } = {
 
     if (offerService.emit) {
       const freshOffer = await Offer.findByPk(offer.provider) as Offer
-      offerService.emit('updated', wrapEvent('BillingPlanSet', freshOffer.toJSON()))
+
+      if (freshOffer) {
+        offerService.emit('updated', wrapEvent('BillingPlanSet', freshOffer.toJSON()))
```
AuHau, line 86: Hmm should we emit `deleted` for when Billing Price is removed? 🤔

src/services/storage/models/billing-plan.model.ts
```diff
@@ -8,23 +8,23 @@ import Rate from '../../rates/rates.model'
 
 @Table({ freezeTableName: true, tableName: 'storage_billing-plan' })
 export default class BillingPlan extends Model {
-  @Column({ ...BigNumberStringType('period') })
+  @Column({ ...BigNumberStringType('period'), allowNull: false })
```
AuHau, line 11: 👍

test/services/storage/processor.spec.ts
```diff
@@ -160,6 +160,23 @@ describe('Storage services: Events Processor', () => {
...
+        // Create new offer and billing plan
+        const offer = await Offer.create({ provider })
+        const billing = await BillingPlan.create({ offerId: offer.provider, period: 99, price: 1, tokenAddress: token, rateId: 'rbtc' })
+        expect(offer).to.be.instanceOf(Offer)
+        expect(billing?.price).to.be.eql(new BigNumber(1))
+        expect(billing?.tokenAddress).to.be.eql(token)
+        expect(billing).to.be.instanceOf(BillingPlan)
+
+        // should remove billing price
+        billingEvent.returnValues.price = 0
```
AuHau, line 173: Hmm sorry did not catch this in the Review where you were adding these tests. 

Can you please create the `billingEvent` in every test? This smells like a problem, when you will forget to change something in some other tests and it will start magically fails. Or move the instantion of the `billingEvent` to `beforeEach` block...

AuHau (changes requested):
Generally LGTM, just few nitpicks and something that I did not notice in earlier reviews.

src/services/storage/handlers/offer.ts
nduchak: Is it already sent by feather service by default?

test/services/storage/processor.spec.ts
nduchak: Done

src/services/storage/handlers/offer.ts
AuHau: It would be if you would use `offerService.remove()` but not when you remove it using the Sequelize's model directly.

nduchak: Yes, you are right. But anyway we do not have BillingPlan service to emit `remove` event for billing plan

nduchak: Do we really need that?

AuHau (approved):
LGTM!

PR #345, opened 2020-10-21
fix: return total staked fiat as 0.00 for empty account

farinalvaro:
Solves the problem of returning "NaN" when trying to get the stakes for an account without funds

### DIFF
--- a/src/services/storage/index.ts
+++ b/src/services/storage/index.ts
@@ -44,9 +44,9 @@ export class StakeService extends Service {
     const sequelize = this.Model.sequelize
 
     const query = getStakesForAccount(sequelize, account)
-    const [{ totalStakedFiat }] = await sequelize.query(query, { type: QueryTypes.SELECT, raw: true })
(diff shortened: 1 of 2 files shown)


julianmrodri (approved):
LGTM

src/services/storage/index.ts
```diff
@@ -44,9 +44,9 @@ export class StakeService extends Service {
     const sequelize = this.Model.sequelize
 
     const query = getStakesForAccount(sequelize, account)
-    const [{ totalStakedFiat }] = await sequelize.query(query, { type: QueryTypes.SELECT, raw: true })
+    const [{ totalStakedFiat }] = (await sequelize.query(query, { type: QueryTypes.SELECT, raw: true }))
```
farinalvaro, line 47:
```suggestion
    const [{ totalStakedFiat }] = await sequelize.query(query, { type: QueryTypes.SELECT, raw: true })
```

nduchak (approved):
LGTM 🥇

PR #363, opened 2020-11-02
chore: add networkId for ganache and staging

julianmrodri:
* Add `networkId` parameter for `ganache` and `staging` config files.


nduchak (approved):
LGTM 🥇

PR #365, opened 2020-11-03
docs: enable lead maintainer for the repo

AuHau:


CODEOWNERS
```diff
@@ -0,0 +1 @@
+* adam@uhlir.dev
```
jurajpiar, line 1: shouldn't you use your iovlabs account here?

nduchak (approved):

PR #367, opened 2020-11-05
fix: make total-capacity(offer) and size(agreement) number type

nduchak:


src/services/storage/models/offer.model.ts
```diff
@@ -49,7 +48,7 @@ export default class Offer extends Model {
 
   @Column(DataType.VIRTUAL)
   get availableCapacity (): BigNumber {
-    return this.totalCapacity.minus(this.utilizedCapacity)
+    return new BigNumber(this.totalCapacity).minus(this.utilizedCapacity)
```
julianmrodri, line 51: I think this is a bit hacky. Because we are keeping `utilizedCapacity` as BigNumber but it will even smaller than `totalCapacity`. I suggest also making `utilizedCapacity` a number, and then here you would need to do normal subtraction `this.totalCapacity - this.utilizedCapacity`).

src/services/storage/handlers/offer.ts
```diff
@@ -72,6 +72,7 @@ const handlers: { [key: string]: Function } = {
     const flag = firstMsg.substring(2, 4)
 
     if (flag === '01') { // PeerId definition
+      const oldPeerId = offer.peerId
```
julianmrodri, line 75: I would rename to `previousPeerId` but its just a personal preference than "old" in this situation.

julianmrodri (approved):
LGTM. Made two small comments which we might want to address.

src/services/storage/handlers/offer.ts
nduchak: Done

src/services/storage/models/offer.model.ts
nduchak: Done

src/services/storage/models/offer.model.ts
```diff
@@ -28,8 +27,8 @@ export default class Offer extends Model {
   @Column({ primaryKey: true, type: DataType.STRING(64) })
   provider!: string
 
-  @Column({ ...BigNumberStringType('totalCapacity') })
-  totalCapacity!: BigNumber
+  @Column
+  totalCapacity!: number
```
AuHau, line 31: Hmmm I assume this is because of the sorting right? 

Then I guess we should do some size overflow guards here... @julianmrodri do we have defined "max capacity value" for Offer that we allow or just utilizing the full types range? 

We have:
 - in contract, this is `uint128`
 - JS supports [58 bits](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)
 - what about the DB here?

Should we maybe downsize the contract to `uint64`? With `uint64` we are supposed to get 18x10e18, so since we have the size in Mb, I would argue supporting Exa bytes would be enough? 😅

AuHau (changes requested):

src/services/storage/models/offer.model.ts
julianmrodri: Sqlite create san `INTEGER` I believe its 64-bit. We can downsize the contract probably just in case.

AuHau: Ok, but we are still hitting the JS limitation then...

src/services/storage/handlers/offer.ts
```diff
@@ -72,6 +72,7 @@ const handlers: { [key: string]: Function } = {
     const flag = firstMsg.substring(2, 4)
 
     if (flag === '01') { // PeerId definition
+      const previousPeerId = offer.peerId
       offer.peerId = decodeByteArray([`0x${firstMsg.substring(4)}`, ...restMsg])
 
       await offer.save()
```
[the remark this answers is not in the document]
nduchak: Agree, this looks better

src/services/storage/models/offer.model.ts
nduchak: this bit tricky, as sequelize not support BigNumber or some lib for handling big numbers, we can not store 64-bit number using js

nduchak: The best way to resolve this will be to store it as a string and prepare a custom query with casting to number for retrieving min/max value

julianmrodri: Maybe we can validate the max capacity for Offers to the max number JS can handle. This can be done both in UI and contract if we want.

julianmrodri: @nduchak im fine with that. Would keep us safer. Is that something you can sneak at tackle?

nduchak: Yes, I guess I can tackle both parts UI and cache

nduchak: Not sure if I had enough time for that now

julianmrodri: @AuHau are ytou ok with custom query approach?

AuHau: Lets discuss this after daily

nduchak: Discussion notes: change this column to `BigInt` sequelize datatype and prepare a custom getter/setter for casting to`BigNumber` in JS

nduchak: Done

src/sequelize.ts
```diff
@@ -58,6 +54,33 @@ export function BigNumberStringType (propName: string): Partial<ModelAttributeCo
   }
 }
 
+/**
+ * consider that the field will be stored as string in the data base,
+ * so you not be able to use number comparision when querying
+ * BigNumberStringType for sequelize models
```
AuHau, line 60:
```suggestion
 * Consider that the field will be stored as a string in the database,
 * so you won't be able to use number comparison when querying and also
 * data are ordered lexicographically so most probably ordering won't work
 * as you would expect.
```

src/sequelize.ts
```diff
@@ -58,6 +54,33 @@ export function BigNumberStringType (propName: string): Partial<ModelAttributeCo
...
+export function BigNumberStringType (propName: string): Partial<ModelAttributeColumnOptions> {
+  return {
+    type: DataType.STRING(),
+    ...bigNumberGetterSetter(propName)
+  }
+}
+
+/**
+ * consider that the field will be stored as number in the data base,
+ * BigNumberStringType for sequelize models
```
AuHau, line 73:
```suggestion
 * Consider that the field will be stored as BigInt in the database. 
 * Based on the database engine this might have limited precision. 
 * Using comparison and ordering is supported.
```

AuHau (approved):
Few docs tweaks but otherwise LGTM!

PR #378, opened 2020-11-09
feat: adjust room topic computation

nduchak:


AuHau (approved):
LGTM, but lets hold with merge until we have this fix everywhere...

PR #381, opened 2020-11-12
fix(dev): re-adds the watch arg to bin script

### LINKED ISSUE #380: Re-enable hot reload --watch
Investigate why it is not working in the Cache. Maybe we can add a separate command.

jurajpiar:
Puts the `--watch` argument back to the bin script in _package.json_

Resolves #380


AuHau:
Ufff I am a bit against this. Can you create `bin:watch` script which gonna watch it, please?

AuHau (changes requested):
Ufff I am a bit against this. Can you create bin:watch script which gonna watch it, please?

AuHau (approved):
LGTM, thanks!

PR #384, opened 2020-11-12
fix(storage): makes periodsSinceLastPayout 0 if not positive

farinalvaro:
It seems that `lastPayout.getTime()` is 5 minutes ahead of `Date.now()`. This fix defaults to 0 when that time is bigger than now

### DIFF
--- a/src/services/storage/models/agreement.model.ts
+++ b/src/services/storage/models/agreement.model.ts
@@ -57,7 +57,11 @@ export default class Agreement extends Model {
     // Date.now = ms
     // this.lastPayout.getTime = ms
     // this.billingPeriod = seconds ==> * 1000
-    const timePassedMs = Math.max(Date.now() - this.lastPayout.getTime(), 0)
+    const timePassedMs = Date.now() - this.lastPayout.getTime()
+    if (timePassedMs < 0) {
+      // We cannot fully rely on blockchain time, as it can be a little bit ahead of real time
+      return new BigNumber(0)
+    }
     const periods = new BigNumber(timePassedMs).div(this.billingPeriod.times(1000))
     return floor
       ? bnFloor(periods)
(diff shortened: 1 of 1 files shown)


julianmrodri:
Just to give some more context we detected they may be micro differences between the `timestamp` in ganache vs the Cache, and that resulted in some negative value during a short period of time. This is a safe check to ensure we handle those micro differences for the UI.

src/services/storage/models/agreement.model.ts
```diff
@@ -57,7 +57,8 @@ export default class Agreement extends Model {
     // Date.now = ms
     // this.lastPayout.getTime = ms
     // this.billingPeriod = seconds ==> * 1000
-    const periods = new BigNumber(Date.now() - this.lastPayout.getTime()).div(this.billingPeriod.times(1000))
+    const timePassedMs = Math.max(Date.now() - this.lastPayout.getTime(), 0)
```
nduchak, line 60: I prefer to separate this:
```
if (Date.now() - this.lastPayout.getTime() < 0) {
  // We ca not fully rely on blockchain time, as it's can be little bit ahead a real time
  return new BigNumber(0)
}
const periods = new BigNumber(Date.now() - this.lastPayout.getTime()).div(this.billingPeriod.times(1000))
return floor
      ? bnFloor(periods)
      : periods

```

nduchak, line 60: I prefer to separate this:
```
if (Date.now() - this.lastPayout.getTime() < 0) {
  // We ca not fully rely on blockchain time, as it's can be little bit ahead a real time
  return new BigNumber(0)
}
const periods = new BigNumber(Date.now() - this.lastPayout.getTime()).div(this.billingPeriod.times(1000))
return floor
      ? bnFloor(periods)
      : periods

```

nduchak (approved):
LGTM

julianmrodri (approved):
LGTM

AuHau (approved):

PR #385, opened 2020-11-13
chore: new contract and config

julianmrodri:
* New contract version `v6`
* Updates addresses for ganache and staging.


nduchak (approved):

AuHau:
That fix is thanks to @nduchak who found it out, I was stuck 😅

PR #387, opened 2020-11-13
chore: update docker entrypoint for migrations

AuHau:


Dockerfile
```diff
@@ -26,8 +26,6 @@ RUN sed -i 's#"./src/cli"#"./lib/cli"#g' package.json
 
 EXPOSE 3030
 
-ENV RIFM_PRECACHE_TARGETS all
```
nduchak, line 29: What about precache, when it will be running then?

bin/entrypoint
```diff
@@ -1,5 +1,4 @@
 #!/bin/sh
 
-./bin/run db-sync "$@"
-./bin/run precache "$@" $RIFM_PRECACHE_TARGETS
```
nduchak, line 4: What about precache, when it will be running then?

AuHau: No need. `start` will continue where it ended. I assume that the updates are fairly quick so it will be only few blocks behind. 

It was a design decision to require run `precache` in order to split the long running fetching of data from the start of blockchain from actually being ready to serve data to user... Technically the `precache` did not have to be there as `start` would just start fetching data from "latest block" (which would be the genesis block in that case)...

AuHau: It will require to run `precache` on the first deployment with the migrations though...

nduchak: Clear

nduchak (approved):
LGTM

PR #401, opened 2020-11-18
chore: topics for RNS events

### LINKED ISSUE #244: Add topics for RNS
With the PR https://github.com/rsksmart/rif-marketplace-cache/pull/243 there is new support for Event's topics. This should filter the events based on the Event's signature to only those we are interested directly on the RSK node and not fetch everything. 

I don't know what are the RNS's events signatures so please add those. You can see the example https://github.com/rsksmart/rif-marketplace-cache/pull/243/files#diff-0b4d6dead760d533c5ef6036feb32d4bR47

julianmrodri:
* Adds Topics for RNS to `default` configuration
* `ExpirationChanged(uint256,uint)` is defined in the contract using `uint256` and `uint` for each parameter. Although these represent the same data type I kept the same naming as the Contract. But I wasnt sure if here we should keep it like this or use `uint256` in both params.  Here is the definition:  https://github.com/rnsdomains/rns-rskregistrar/blob/94b7226c29a2c76cb5f234c86a1a0fc8ff600aba/contracts/NodeOwner.sol#L17

Closes #244

### DIFF
--- a/config/default.json5
+++ b/config/default.json5
@@ -150,7 +150,7 @@
         topics: [
           [ // It needs to be a "double array" because that represents an "or" of the topics and not "and"
             'Transfer(address,address,uint256)',
-            'ExpirationChanged(uint256,uint)',
+            'ExpirationChanged(uint256,uint256)',
             'Approval(address,address,uint256)'
           ]
         ],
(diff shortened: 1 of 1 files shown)


AuHau (approved):
LGTM, yeah topics should be using fully classified types. So no `uint` but `uint256`, I think 😅

nduchak (approved):
LGTM

PR #402, opened 2020-11-18
feat: add check for eth provider availability

nduchak:


AuHau (approved):
LGTM, thanks for making this! 👍

PR #414, opened 2020-11-23
feat(storage): adds ability to filter by availableCapacity

farinalvaro:
- Creates new service to fetch min and max values of offers availableCapacity
- Adds availableCapacity filter

Related to https://github.com/rsksmart/rif-marketplace-ui/issues/460


src/services/storage/hooks/offers.hooks.ts
```diff
@@ -101,7 +128,8 @@ export default {
             attributes: {
               include: [
                 [getBillingPriceAvgQuery(sequelize, 'usd'), 'avgBillingPrice'],
-                [getStakesAggregateQuery(sequelize, 'usd'), 'totalStakedUSD']
+                [getStakesAggregateQuery(sequelize, 'usd'), 'totalStakedUSD'],
+                [getAvailableCapacityQuery(), 'availableCapacity']
```
AuHau, line 132: Shouldn't be this added only when `availableCapacity`?

src/services/storage/models/offer.model.ts
```diff
@@ -112,3 +107,23 @@ export function getBillingPriceAvgQuery (
...
+ * This function generates nested query for available capacity for offer
+ */
+export function getAvailableCapacityQuery (): Literal {
+  return literal(`
+    totalCapacity - COALESCE(
+      (SELECT
+        CAST(
+          SUM(
+            "storage_agreement"."size"
+            ) as INTEGER
```
AuHau, line 121: Hmm the `INTEGER` part is to remove decimals? I mean I assume that the `size` is a number already anyway no? Otherwise you could not SUM it...

AuHau (changes requested):
Few questions, but mainly please add documentation about these changes to REAMDE ;-)

src/services/storage/models/offer.model.ts
nduchak: True, this can be removed

nduchak: actually, it's better to cast to a string, to make sure that we don't have a problem with big numbers in JS

src/services/storage/hooks/offers.hooks.ts
farinalvaro: Hmm I don't think so because this is [replacing the virtual column `availableCapacity`](https://github.com/rsksmart/rif-marketplace-cache/pull/414/files#diff-494e5b8a4ef33c59f54a87ed47867e8cf912680e6a00339a3af0cf2514ed327dL50) that was always provided when fetching offers. So we should keep providing this value always, right?

AuHau: Sounds good!

src/services/storage/models/offer.model.ts
farinalvaro: Done, thank you guys

src/services/storage/models/offer.model.ts
```diff
@@ -112,3 +107,23 @@ export function getBillingPriceAvgQuery (
...
+ * This function generates nested query for available capacity for offer
+ */
+export function getAvailableCapacityQuery (): Literal {
+  return literal(`
+    totalCapacity - COALESCE(
+      (SELECT
+        CAST(
+          SUM(
+            "storage_agreement"."size"
+            ) as STRING
```
nduchak, line 121: Hmm... You should wrap the whole thing with `CAST as String` instead of wrapping just availableCapacity, as we had totalCapacity(Number) - availableCapacity(String)

test/unit/services/storage/models.spec.ts
```diff
@@ -332,6 +333,41 @@ describe('Models', () => {
       ]
       expect(offers).to.be.deep.equal(expectedRes)
     })
+    it('should aggregate available size for offers', async () => {
```
nduchak, line 336: Also, can you please add test for retrieving min/max value for `availableCapacity`

nduchak (changes requested):

AuHau (approved):
Please addres Naz's comments and then it LGTM!

src/services/storage/utils.ts
```diff
@@ -38,3 +38,25 @@ export function getAvgMinMaxBillingPriceQuery (minMax: MinMax): string {
...
+ * Get query for generating minimum or maximum available capacity
+ * @param minMax
+ */
+export function getMinMaxAvailableCapacityQuery (minMax: MinMax): string {
+  return `
+  SELECT
+      CAST(
+        totalCapacity - SUM(
+          COALESCE("storage_agreement"."size", 0)
+          ) as STRING
```
nduchak, line 52: Can you please remove casting to String here, and instead wrap availableCapacity field

src/services/storage/models/offer.model.ts
```diff
@@ -112,3 +107,25 @@ export function getBillingPriceAvgQuery (
...
+ */
+export function getAvailableCapacityQuery (): Literal {
+  return literal(`
+    CAST(
+      totalCapacity - COALESCE(
+        (SELECT
+          CAST(
+            SUM(
+              "storage_agreement"."size"
+              ) as STRING
```
nduchak, line 122: Please remove cast to string here

nduchak (changes requested):

nduchak (approved):
Nice job 👍 
Just move the test

test/unit/services/storage/models.spec.ts
```diff
@@ -343,7 +343,7 @@ describe('Models', () => {
         { provider: 'provider2', totalCapacity: 100, peerId: '2' }
       ])
       await Agreement.bulkCreate([
-        { agreementReference: '123', size: 10.0000000000000001, offerId: 'provider1' },
```
farinalvaro: I removed the decimals as we are not storing decimals in the DB so was not a real test

PR #418, opened 2020-11-25
fix(core): logs colors handling

### LINKED ISSUE #417: Support no colors output
DevOps are collecting logs from Container's STDOUT. We should enable configure no colors output when env variable `LOG_NO_COLORS` is set.

AuHau:
Resolves #417 

Uses the `supportsColor()` and also allows env. variable override.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,7 +1,8 @@
 import { createLogger, format, transports, addColors, Logger as RealLogger } from 'winston'
 import * as Transport from 'winston-transport'
 
-import colors from 'colors'
+import colors from 'colors/safe'
+import { supportsColor } from 'colors/lib/system/supports-colors'
 import config from 'config'
 
 import { Logger } from './definitions'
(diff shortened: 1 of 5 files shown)


nduchak (approved):

src/logger.ts
```diff
@@ -132,7 +133,7 @@ function initLogging (): void {
       upperCaseLevel(),
       // format.padLevels(),
       format.timestamp({ format: 'DD/MM hh:mm:ss' }),
-      format.colorize(),
+      !supportsColor() || process.env.LOG_NO_COLORS === 'true' ? format(i => i)() : format.colorize(),
```
lucasvuotto, line 136: The chunk below has a hardcoded `colors.grey`, which I think is not going to be affected by this changes.

lucasvuotto: And from my probably limited understanding of how color is being handled by the library, https://github.com/winstonjs/logform/blob/master/colorize.js#L9 enables colors under any circumstance, offering no way of make `colors.grey` a no-op.

AuHau: Hmm thanks for pointing this out, I missed that one 😅

In the end I choose instead to put `uncolorize()` there as we have some colored input coming in directly from our code, so this might not be as optimal, but will work.

PR #420, opened 2020-11-26
fix(storage): converts token address to lower case

jurajpiar:
We should be storing token addresses with lower case.

### DIFF
--- a/src/services/storage/handlers/offer.ts
+++ b/src/services/storage/handlers/offer.ts
@@ -95,7 +95,7 @@ const handlers: { [key: string]: Function } = {
     }
   },
   async BillingPlanSet ({ returnValues: { period, price, token } }: EventData, offer: Offer, offerService: OfferService): Promise<void> {
-    await updatePrices(offer, period, price, token)
+    await updatePrices(offer, period, price, token.toLowerCase())
 
     if (offerService.emit) {
       const freshOffer = await Offer.findByPk(offer.provider) as Offer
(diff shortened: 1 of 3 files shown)


src/services/storage/handlers/offer.ts
```diff
@@ -95,7 +95,7 @@ const handlers: { [key: string]: Function } = {
     }
   },
   async BillingPlanSet ({ returnValues: { period, price, token } }: EventData, offer: Offer, offerService: OfferService): Promise<void> {
-    await updatePrices(offer, period, price, token)
+    await updatePrices(offer, period, price, token.toLowerCase())
```
nduchak, line 98: hmm.. we have an inconsistency in the cache and contract then, should we also lower case token address when creating a transaction to the storage contract?

jurajpiar: No because we should allow users to use the contract with either format.

AuHau (approved):
LGTM

AuHau:
If you would feel fancy, you could write a closure that the events pass through before reaching the handler and you could specify which properties should be lowercased like:
```
toLowerCaseAttributes(..., ['tokenAddress', ...])
```

nduchak (approved):

jurajpiar:
This is being addressed [elsewhere](https://github.com/rsksmart/rif-marketplace-cache/pull/423/files#diff-cabdeead4c1efc53f99c6dc1689e8d67493999c352724eaada23f25eccac5bd1R107).

PR #423, opened 2020-11-29
fix: handle libp2p messages which comes before agreement event

nduchak:
implement retry with timeout mechanism for async function

### DIFF
--- a/src/communication/handlers.ts
+++ b/src/communication/handlers.ts
@@ -5,6 +5,7 @@ import Agreement from '../services/storage/models/agreement.model'
 import { NotificationService } from '../notification'
 import { loggingFactory } from '../logger'
 import { asyncRetry } from '../utils'
+import { AsyncRetryError } from '../errors'
 
 const logger = loggingFactory('communication:handlers')
 
@@ -70,11 +71,11 @@ function buildNotification (agreement: Agreement, message: CommsMessage<CommsPay
   }
 }
 
-async function getAgreementWithRetry (
+function getAgreementWithRetry (
   agreementRef: string,
   retriesCount: number,
   retryInterval: number
-): Promise<Agreement | undefined> {
+): Promise<Agreement | void> {
   function getAgreement (): Promise<Agreement> {
     logger.info('Trying to get agreement ' + agreementRef)
     return Agreement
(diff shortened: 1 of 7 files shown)


src/communication/handlers.ts
```diff
@@ -69,6 +70,34 @@ function buildNotification (agreement: Agreement, message: CommsMessage<CommsPay
...
+        }
+        return agreement
+      })
+  }
+  return await asyncRetry<Agreement>(
+    getAgreement,
+    retriesCount,
+    retryInterval
+  ).catch(e => {
+    if (e.message.indexOf('Max retries reached for function') !== -1) {
```
AuHau, line 94: Uff this is not really good way to do this... Error's message should not be a way how you assert errors if you have any other way to do so and since this is our code, you can do it better ;-)

Messages are supposed to be for human not for code ;-) What if you change to wordings of the messages/fix a typo? Everything will brake...

Please create custom error like for example [here](https://github.com/rsksmart/rif-storage-pinner/blob/master/src/errors.ts#L38) and then use it like [here](https://github.com/rsksmart/rif-storage-pinner/blob/5590697fd7cd224b9a430bbcd2b3232066ebd389/src/gc.ts#L60)

src/communication/handlers.ts
```diff
@@ -69,6 +70,34 @@ function buildNotification (agreement: Agreement, message: CommsMessage<CommsPay
...
+      return undefined
```
AuHau, line 95:
```suggestion
      return
```

Undefined is not necessary...

AuHau (changes requested):
Pretty much LGTM! The only main thing is regarding how you handle/throw errors.

Also, I am thinking if maybe the message should not be logged before it starts retrying? In case Cache would terminate for some reason and the message would be lost, we could have at least have it recorded in logs?

test/unit/services/storage/communication.spec.ts
```diff
@@ -74,6 +80,38 @@ describe('Communication', function () {
     expect(notifications[0].type).to.be.eql(NotificationType.STORAGE)
     expect(notifications[0].payload).to.be.eql({ ...message.payload, code: message.code })
   })
+  it('Should create notification if agreement comes later on', async () => {
```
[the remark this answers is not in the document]
nduchak: Done

src/communication/handlers.ts
nduchak: Done

AuHau (approved):
LGTM! Thanks!

PR #431, opened 2020-12-02
refactor(storage): renames nonactive to inactive

jurajpiar:
non-active is not a word plus it's an awkward object key


src/services/storage/hooks/offers.hooks.ts
```diff
@@ -93,8 +93,8 @@ export default {
           nest: true
         }
 
-        if (context.params?.query?.['non-active']) {
-          delete context.params?.query?.['non-active']
+        if (context.params?.query?.['inactive']) {
+          delete context.params?.query?.['inactive']
```
jurajpiar, line 97: In fact why not just ask whether it is active rather than having to invent another word?

jurajpiar: How about this?
```
        if (context.params?.query?.['active']) {
          context.params.sequelize.scope = 'active'
        }
        delete context.params?.query?.['inactive']
```
I'm thinking the query param should be deleted either way actually, since we are interested in the scope not a field

PR #451, opened 2020-12-11
fix: lower case migration error

nduchak:


AuHau (approved):
Does this affect anything?

nduchak:
No, as we don't have this migration yet on staging.

PR #493, opened 2021-01-12
hotfix(rns): rollback of rns exp. date changes 

jurajpiar:


src/migrations/scripts/1609770161070-normalises_rns_expiration.ts
```diff
@@ -1,41 +0,0 @@
-import { QueryInterface } from 'sequelize'
```
AuHau, line 1: Ufff, you should not remove migrations!!! Once migration is created it should never be modified/deleted unless there are some bugs in it (eq. it is crashing and can not be run).

I guess this time it is alright as this change was not yet deployed. Anyhow you should *warn everybody at least that they will have to remove DB and precache*...

AuHau (approved):
So you verified with RNS team that the order of the events is by design like we discovered yesterday?

src/migrations/scripts/1609770161070-normalises_rns_expiration.ts
jurajpiar: In this case I really saw no benefit in that. As to telling everyone, we always test with / use clean db as far as I'm aware...

PR #502, opened 2021-01-18
chore: updated tasegir

nduchak:


src/rates/update.ts
```diff
@@ -63,7 +63,7 @@ export async function updater (): Promise<void> {
           throw Error('Unsupported fiat')
         }
 
-        // @ts-ignore
+        // @ts-ignore: not typed
```
AuHau, line 66: Heh lazy you 🤪 This does meet the requirement, but not the reason for that requirement 😉 Some better explanation would be better...

src/sequelize.ts
```diff
@@ -29,7 +29,7 @@ export function sequelizeFactory (): Sequelize {
       return filename.substring(0, filename.indexOf('.model')) === member.toLowerCase()
     },
     logging: (msg) => logger.debug(formatLogs(msg)),
-    // @ts-ignore
+    // @ts-ignore: Not typed
```
AuHau, line 32:
```suggestion
    // @ts-ignore: Incorrectly typed in the sequelize library
```

src/services/storage/handlers/offer.ts
```diff
@@ -113,7 +113,7 @@ const handler: Handler<StorageOfferEvents, StorageServices> = {
     const provider = event.returnValues.provider
 
     // TODO: Ignored until https://github.com/sequelize/sequelize/pull/11924
-    // @ts-ignore
+    // @ts-ignore: issue ^
```
AuHau, line 116:
```suggestion
    // @ts-ignore: TODO: Ignored until https://github.com/sequelize/sequelize/pull/11924 
```

src/services/storage/handlers/offer.ts
```diff
@@ -113,7 +113,7 @@ const handler: Handler<StorageOfferEvents, StorageServices> = {
     const provider = event.returnValues.provider
 
     // TODO: Ignored until https://github.com/sequelize/sequelize/pull/11924
```
AuHau, line 115:
```suggestion
```

AuHau (changes requested):
There is a reason for the lint rule for the `@ts-*` directives, putting everywhere "not typed" is meeting the requirement but not the reason for it. If you won't do it now, you will never do it. I can accept few places where are you not sure for it, but putting it everywhere is not a really a good idea. At least comment the ones in `/src`, we can ignore the test's ones...

src/rates/update.ts
nduchak: Hehe, agree. 
Will fix it.

nduchak: Done

AuHau (approved):
Sweet! LGTM now! Thanks for the explanations.

PR #516, opened 2021-01-27
feat(rns): allows sorting and filtering by fiat

### LINKED ISSUE #663: Sorting in RNS by Price should use USD
When sorting in the RNS Buy Screen by price, we should be using the **USD** value now that we have multicurrency. It is currently ordering by number but depending on the Currency prices will differ. See example below where it puts as first row the highest amount, but should be the last one. We need to use USD values when sorting by Price.

 image

jurajpiar:
- copies and adapts usdPrice calculation from storage to rns to allow sorting of offers by fiat instead of crypto

- removes the feather pagination of rns offers as this caused the issues in execution of custom literal attributes (https://github.com/rsksmart/rif-marketplace-cache/issues/528).

Resolves https://github.com/rsksmart/rif-marketplace-ui/issues/663
Followed by: https://github.com/rsksmart/rif-marketplace-ui/issues/664
for [RMKT#523](https://rsklabs.atlassian.net/browse/RMKT-523)


src/migrations/scripts/1597921613888-initial.ts
```diff
@@ -604,7 +604,19 @@ const migrationCommands = function (transaction: any): Commands {
           creationDate: {
             type: Sequelize.DATE,
             field: 'creationDate'
-          }
+          },
```
nduchak, line 607: This should be new migration

jurajpiar: This has been addressed in the description. In fact, this is the reason it’s still a draft.

nduchak (approved):
LGTM

PR #548, opened 2021-03-01
chore: fix potential log injection

artem-brazhnikov:
Affected Assets

https://github.com/rsksmart/rif-marketplace-cache/releases/tag/v1.3.2

Description
Log Injection (also known as Log Forging) is a vulnerability where an attacker can forge new log entries, due to a bad input sanitisation by the application.

Remediation
All external data should be sanitised before sending it to logs.

When using a simple one line per entry format for the logs, new lines characters should be removed or escaped so that an attacker cannot create new entries.

References
https://owasp.org/www-community/attacks/Log_Injection

https://cheatsheetseries.owasp.org/cheatsheets/Logging_Cheat_Sheet.html

https://thesecurityvault.com/appsec/log-forging-vulnerability-and-how-to-fix-it/

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -143,11 +143,12 @@ function initLogging (): void {
   })
 }
 
-export function formatLogMessage(info: any): string {
+export function formatLogMessage (info: any): string {
   let message: string
   const { service, ...rest } = info.metadata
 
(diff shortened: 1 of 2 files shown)


nduchak (approved):
LGTM 🥇 
Please fix Linter

nduchak (approved):

PR #572, opened 2021-03-11
feat: base integration of triggers

nduchak:
This PR contains a basic integration of Triggers service:
- service structure
- service precache
- adjust CLI
- add configuration
- migrations
- event handling and DB models

### DIFF
--- a/src/services/triggers/handlers/provider.ts
+++ b/src/services/triggers/handlers/provider.ts
@@ -28,8 +28,8 @@ const handlers = {
   }
 }
 
-function isValidEvent (value: string): value is keyof typeof handlers {
-  return value in handlers
+function isValidEvent (eventName: string): eventName is keyof typeof handlers {
+  return eventName in handlers
 }
 
 const handler: Handler<NotificationManagerEvents, TriggersServices> = {
--- a/src/services/triggers/handlers/provider.ts
+++ b/src/services/triggers/handlers/provider.ts
@@ -1,0 +1,46 @@
+import { Eth } from 'web3-eth'
+import { ProviderRegistered } from '@rsksmart/rif-marketplace-notifications/types/web3-v1-contracts/NotificationsManager'
+
+import { loggingFactory } from '../../../logger'
+import { Handler, NotificationManagerEvents } from '../../../definitions'
+import { wrapEvent } from '../../../utils'
+
+import { TriggersServices } from '../index'
+import ProviderModel from '../models/provider.model'
+
+const logger = loggingFactory('triggers:handler:provider')
+
+const handlers = {
+  async ProviderRegistered (event: ProviderRegistered, { providerService }: TriggersServices): Promise<void> {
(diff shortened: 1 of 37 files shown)


src/services/triggers/handlers/provider.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    if (providerService.emit) providerService.emit('created', wrapEvent('ProviderRegistered', { provider, url }))
+    logger.info(`Created new Provider with address ${provider} and url  ${url}`)
+  }
+}
+
+function isValidEvent (value: string): value is keyof typeof handlers {
+  return value in handlers
+}
```
jurajpiar:
```suggestion
function isValidEvent (eventName: string): eventName is keyof typeof handlers {
  return eventName in handlers
}
```

jurajpiar (approved):
LGTM
I'm guessing you are going to add the other fields to the triggers model, right?
Also, perhaps we should agree upon the name for this service, since the dev project calls it notifications, the cache triggers (personally I'd vote for it) and UI notifier (since notifications is already a svc for marketplace notifs).
One thing I noticed when running it is that the precache doesn't process triggers when run with the `all` arg.

Couldn't test whether it works with the UI for some problems with the contract addresses I have (locally I suspect).

0xartem (approved):
In terms of coding I don't have comments, looks good. But I have troubles to fit the whole picture of interactions between different modules inside of my head, so I'm not completely clear on what to recommend as improvements. Maybe next sprint I will be able to spend more time with the code to give you more valuable feedback.

src/services/triggers/handlers/provider.ts
nduchak: Done

nduchak:
> LGTM
> I'm guessing you are going to add the other fields to the triggers model, right?

I will add another model for `Subscription plans` when do integration with the notifier service

> One thing I noticed when running it is that the precache doesn't process triggers when run with the `all` arg.

Hmm... Works for me. Probably some configuration issue. Try to use ganache config (NODE_ENV=ganache)

> Couldn't test whether it works with the UI for some problems with the contract addresses I have (locally I suspect).

This version probably will not work with UI as it doesn't have the required data for the provider table

jurajpiar:
> I will add another model for Subscription plans when do integration with the notifier service

Sure, though it would have been easier for me if it did have the plans already. But, I understand if the reason for delaying this was that you don't know exactly the structure you'll be getting from the notifier service.

> Hmm... Works for me. Probably some configuration issue. Try to use ganache config (NODE_ENV=ganache)

Probably doing something else wrong then. I do use NODE_ENV=ganache.

> This version probably will not work with UI as it doesn't have the required data for the provider table

That is not what I meant. Sorry, should have been clearer. When I created an offer (and forwarded bc), the cache did not capture this event from the BC. I've done this after making sure both cache and ui have the same (and the correct) SC addresses.

PR #584, opened 2021-03-18
feat(triggers): integrates notifier svc

jurajpiar:
- integrates notifier service
- adds subscription plan models


src/services/triggers/hooks/providers.hook.ts
```diff
@@ -26,8 +63,8 @@ export default {
     all: [],
     find: [],
     get: [],
-    create: [],
-    update: [],
+    create: [updatePlans],
```
nduchak: We should move this update from hook directly to the event handler because when we do precache we use service directly without hooks, in this case, we do not retrieve plans

nduchak (changes requested):
Nice job!!!
Just a small comment for precache process

src/services/triggers/hooks/providers.hook.ts
jurajpiar: Right, yea. Makes sense, though that means either changing the handler's signature or disabling db transaction, because in order to create the transaction you need the sequelize object. Or is it possible to get it some other way?
Alternatively, we could add a hook to pre-cache. hm?

jurajpiar: Alternatively, we could add a hook to precache?

jurajpiar: The model instance contains a reference to the sequelize instance.

nduchak (approved):
🥇

src/services/triggers/update.ts
```diff
@@ -0,0 +1,144 @@
...
+export async function updater (sequelize: Sequelize, url?: string): Promise<void> {
+  logger.debug('Acquiring lock for update')
+  await semaphore.acquire()
+
+  try {
+    logger.info('Fetching notifications providers')
+    const options = url && { where: { url } }
+    const providers = await ProviderModel.findAll({ ...options })
+
+    providers.forEach(async (provider) => await updateProvider(provider, sequelize))
```
nduchak: I think we should wait until all providers will be processed because then `semaphore` thing makes no sense

src/services/triggers/hooks/providers.hook.ts
```diff
@@ -1,6 +1,9 @@
 import { HookContext } from '@feathersjs/feathers'
 import { disallow } from 'feathers-hooks-common'
-import { lowerCaseAddressesQueryParamsHook } from '../../utils'
+import PlanModel from '../models/plan.model'
+import TriggersChannelModel from '../models/triggersChannel.model'
+import PriceModel from '../models/price.model'
+// import { lowerCaseAddressesQueryParamsHook } from '../../utils'
```
0xartem, line 6: do we need to keep those comments? you may delete if not

0xartem (approved):
well done!

PR #592, opened 2021-03-22
feat: add script for updating config with precache starting block

nduchak:
Set starting block as `top_block_number - 50`

### DIFF
--- a/scripts/update-precache-blocks.js
+++ b/scripts/update-precache-blocks.js
@@ -22,8 +22,8 @@ const SERVICES = {
 const configFileName = `${NODE_ENV}.json5`
 const pathToConfigFile = path.resolve(__dirname, '../config', configFileName)
 
-const isExist = fs.existsSync(pathToConfigFile)
-if (!isExist) {
+const doesExist = fs.existsSync(pathToConfigFile)
+if (!doesExist) {
   throw new Error(`Config file not found on path = ${pathToConfigFile}`)
 }
 const fileAsString = fs.readFileSync(pathToConfigFile).toString()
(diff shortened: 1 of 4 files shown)


scripts/update-precache-blocks.js
```diff
@@ -0,0 +1,63 @@
...
+const SERVICES = {
+  rns: RNS_CONTRACTS,
+  storage: STORAGE_CONTRACTS,
+  triggers: TRIGGERS_CONTRACTS
+}
+
+const configFileName = `${NODE_ENV}.json5`
+const pathToConfigFile = path.resolve(__dirname, '../config', configFileName)
+
+const isExist = fs.existsSync(pathToConfigFile)
```
0xartem: minor naming: `exists` or `doesExist`

scripts/update-precache-blocks.js
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable */
```
0xartem, line 1: this script will be executed by the devops in their infra?

0xartem (approved):
Good job

scripts/update-precache-blocks.js
nduchak: Yes, probably it will be executed inside the cache docker container

nduchak: Now, I am thinking about defining another strategy for deployment from scratch. There we should define starting blocks as blocks on which contract was deployed

0xartem (approved):
nice, so basically you query the explorer API for the deployment tx depending on the network. Let's see how we can integrate it on the devops side, the CD part will be interesting.

PR #613, opened 2021-04-09
feat(config): adds storage to mainnet

jurajpiar:
Adds missing storage config for mainnet deployment

### DIFF
--- a/config/rskmainnet.json5
+++ b/config/rskmainnet.json5
@@ -3,7 +3,24 @@
     networkId: 30
   },
   storage: {
-    enabled: false,
+    enabled: true,
+    tokens: {
+      '0x2acc95758f8b5f583470ba265eb685a8f45fc9d5': 'rif'
+    },
+    storageManager: {
+      contractAddress: "0xa18c181229F374b00f126fC952775d41A4F4c40e",
+      eventsEmitter: {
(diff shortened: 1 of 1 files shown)


0xartem (approved):

nduchak:
Just thinking should we increase confirmations to 8? As we had 8 on testnet

0xartem:
> Just thinking should we increase confirmations to 8? As we had 8 on testnet

My opinion is that 4 is more than enough. Unless a transaction is  a million dollars or smth it doesn't make sense to wait until 6 confirmations for bitcoin network for example. I suppose the consensus is the same here so we should be fine. Actually even 3 should be fine. In Ethereum dApps they even wait for just 1 to improve UX (I think so) which works fine at scale with a lot of real users.

nduchak:
Agree, but in our case it was done only because of lot of reorgs, and this confirmation range save us from reorg out of range which will require manual resolution

nduchak (approved):

0xartem:
> Agree, but in our case it was done only because of lot of reorgs, and this confirmation range save us from reorg out of range which will require manual resolution

Ok, I didn't know that. Other services use 4 as well. Let's use the same number across all the services. Whatever it will be.

PR #664, opened 2021-05-13
fix(devops): implement workflow w/o tags - staging

rafaiovlabs:


0xartem (approved):
approved, after we add the full CD flow to the UI (develop->staging->testnet->mainnet), we can duplicate it here

PR #673, opened 2021-05-19
fix(devops): add testnet workflow

rafaiovlabs:


.github/workflows/testnet.yml
```diff
@@ -0,0 +1,38 @@
+---
+name: RIF Marketplace CACHE deploy
```
0xartem:
```suggestion
name: RIF Marketplace Testnet CACHE deploy
```

.github/workflows/testnet.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  push:
+    branches:
+      - testnet
+
+jobs:
+  deploy-staging:
```
0xartem: rename the job name please so it corresponds the env name

0xartem (changes requested):

0xartem (approved):

PR #674, opened 2021-05-20
fix(devops): add mainnet workflow

rafaiovlabs:


.github/workflows/mainnet.yml
```diff
@@ -0,0 +1,38 @@
+---
+name: RIF Marketplace CACHE deploy
```
0xartem: Add a mainnet word here plz

0xartem (changes requested):

.github/workflows/mainnet.yml
rafaiovlabs: Done

PR #688, opened 2021-05-25
fix(notifier): adds channels in subscriptions res

jurajpiar:
Adds channels to the plans attribute in the find hook of the subscriptions ep.

Required by https://github.com/rsksmart/rif-marketplace-ui/pull/851

### DIFF
--- a/config/ganache.json5
+++ b/config/ganache.json5
@@ -13,13 +13,13 @@
       '0x67b5656d60a809915323bf2c40a8bef15a152e3e': 'rif'
     },
     notifierManager: {
-      contractAddress: "0xc34175A79ACca40392bECD22ff10fAeBFE780Ae7",
+      contractAddress: "0xeea2Fc1D255Fd28aA15c6c2324Ad40B03267f9c5",
       eventsEmitter: {
         confirmations: 2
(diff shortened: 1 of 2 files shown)


0xartem (approved):
lgtm

config/ganache.json5
```diff
@@ -13,13 +13,13 @@
...
     notifierManager: {
-      contractAddress: "0xc34175A79ACca40392bECD22ff10fAeBFE780Ae7",
+      contractAddress: "0xeea2Fc1D255Fd28aA15c6c2324Ad40B03267f9c5",
       eventsEmitter: {
         confirmations: 2
       }
     },
     staking: {
-      contractAddress: "0xe97DbD7116D168190F8A6E7beB1092c103c53a12",
+      contractAddress: "0xc34175A79ACca40392bECD22ff10fAeBFE780Ae7",
```
itofarina, line 22: I think the changed addresses are the ones given by dev project. Or did something change? 
If that's not the case we should revert this change.

jurajpiar: These are the addresses that are generated with no blockTime

PR #691, opened 2021-05-27
chore: updates notifier ganache addresses

itofarina:
Based on clean installation of dev project.


vjgene (approved):

jurajpiar (changes requested):
Let's not be hasty. We still don't know which one is correct, do we?

itofarina:
@jurajpiar well these are the addresses we had before and @vjgene and I got the same after a clean installation with latest dev

jurajpiar (approved):

PR #708, opened 2021-06-17
fix(notifier): removes plan duplication

jurajpiar:
When the notifier changes the plan status, the plans are duplicated instead of replaced.
This fixes it.


0xartem (approved):
looks good

itofarina (approved):
👌

PR #709, opened 2021-06-17
feat(notifier): includes url in plan object of subscriptions

itofarina:
Includes `url` property under `plan` object when fetching subscriptions

### DIFF
--- a/src/services/notifier/hooks/subscriptions.hook.ts
+++ b/src/services/notifier/hooks/subscriptions.hook.ts
@@ -28,13 +28,22 @@ export default {
                 {
                   model: NotifierChannelModel,
                   as: 'channels',
-                  attributes: ['name'],
                   required: true
                 }
-              ]
+              ],
+              attributes: {
+                include: [
+                  [
+                    literal('(SELECT url FROM notifier_provider AS provider WHERE SubscriptionModel.providerId = provider.provider)'),
+                    'url'
+                  ]
+                ]
+              }
             }
           ],
-          attributes: { exclude: ['subscriptionPlanId'] }
+          attributes: {
+            exclude: ['subscriptionPlanId']
+          }
         }
       }
     ],


src/services/notifier/hooks/subscriptions.hook.ts
```diff
@@ -28,13 +28,22 @@ export default {
...
-              ]
+              ],
+              attributes: {
+                include: [
+                  [
+                    literal('(SELECT url FROM notifier_provider AS provider WHERE SubscriptionModel.providerId = provider.provider)'),
+                    'url'
+                  ]
+                ]
+              }
```
jurajpiar, line 41: this is unnecessary. just add
```
            {
              model: ProviderModel,
              as: 'provider',
              attributes: ['url']
            }
```
to the list of includes (sibling to the plan object).

src/services/notifier/hooks/subscriptions.hook.ts
```diff
@@ -28,13 +28,22 @@ export default {
                 {
                   model: NotifierChannelModel,
                   as: 'channels',
-                  attributes: ['name'],
```
jurajpiar, line 31: why did we remove the limitation to the name attribute?

jurajpiar (changes requested):
Please describe why and where this is needed, and fix the way the url is included in the output.

src/services/notifier/hooks/subscriptions.hook.ts
itofarina: oops, that was a mistake

src/services/notifier/hooks/subscriptions.hook.ts
itofarina: This is a good suggestion. The problem is that this would nest the url prop under provider object. Which doesn't map with our SubscriptionDTO models in UI. So probably we need to update the models in UI as well.
As this is needed to renew a subscription, I'm resolving this conversation and closing this PR for now as we need to define first which data we need when fetching subscriptions and do properly refactor in UI project.

itofarina:
Closing as we need to define first which models we want to keep in UI and do proper refactor.

PR #718, opened 2021-06-28
fix(notifier): catches errors interacting with provider service

itofarina:
When provider service fails to retrieve the current subscriptions we are not able to update them and we log that error.

_This PR should be squashed_

### DIFF
--- a/src/services/notifier/utils/updaterUtils.ts
+++ b/src/services/notifier/utils/updaterUtils.ts
@@ -57,7 +57,6 @@ export const updateSubscriptionsBy = async (
   const svcProvider = new NotifierSvcProvider({ host, port })
   try {
     const subscriptionsDTO: any[] = await svcProvider.getSubscriptions(consumerAddress, subsHashesToUpdate)
-
     await Promise.all(subscriptionsDTO.map(({
       hash,
       status,
@@ -67,9 +66,9 @@ export const updateSubscriptionsBy = async (
       SubscriptionModel.update(
         { status, paid, notificationBalance },
         { where: { hash } }
-      )
+      ).catch(error => logger.warn(`Unable to update subscription with hash ${hash} in the database`, error))
     ))
   } catch (error) {
-    logger.error('Unable to update subscriptions from notifier service provider', error)
+    logger.warn('Unable to update subscriptions from notifier service provider', error)
   }
 }
--- a/src/services/notifier/utils/updaterUtils.ts
+++ b/src/services/notifier/utils/updaterUtils.ts
@@ -1,8 +1,11 @@
+import { loggingFactory } from '../../../logger'
 import NotifierChannelModel from '../models/notifier-channel.model'
 import PlanModel from '../models/plan.model'
 import PriceModel from '../models/price.model'
 import SubscriptionModel from '../models/subscription.model'
 import { NotifierSvcProvider, SubscriptionPlanDTO } from '../notifierService/provider'
+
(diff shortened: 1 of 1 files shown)


src/services/notifier/utils/updaterUtils.ts
```diff
@@ -52,17 +55,21 @@ export const updateSubscriptionsBy = async (
...
+      paid,
+      notificationBalance
+    }) =>
+      SubscriptionModel.update(
+        { status, paid, notificationBalance },
+        { where: { hash } }
+      )
+    ))
+  } catch (error) {
+    logger.error('Unable to update subscriptions from notifier service provider', error)
```
vjgene: How about using logger.warn instead of error. Since this is just a warning we are giving by logging the error. It would be easier to distinguish also with actual errors.

vjgene (approved):
Looks good. Just one minor suggestion.

src/services/notifier/utils/updaterUtils.ts
```diff
@@ -52,17 +55,21 @@ export const updateSubscriptionsBy = async (
...
+  }
```
jurajpiar:
```suggestion
    await Promise.all(subscriptionsDTO.map(({
      hash,
      status,
      paid,
      notificationBalance
    }) =>
      SubscriptionModel.update(
        { status, paid, notificationBalance },
        { where: { hash } }
      ).catch((error) => {
        logger.error('Unable to update subscriptions from notifier service provider', error)
      })
    ))
```
and remove the try. Otherwise, it might not complete the processing the list.

jurajpiar (changes requested):
consider following

src/services/notifier/utils/updaterUtils.ts
itofarina: Thanks, added the chain catch to the update DB action.
We need to keep the try block to catch possible errors with communication with notifier service. In that case we won't update the DB and log the error.

src/services/notifier/utils/updaterUtils.ts
```diff
@@ -52,17 +55,20 @@ export const updateSubscriptionsBy = async (
...
+      notificationBalance
+    }) =>
+      SubscriptionModel.update(
+        { status, paid, notificationBalance },
+        { where: { hash } }
+      ).catch(error => logger.warn(`Unable to update subscription with hash ${hash} in the database`, error))
+    ))
+  } catch (error) {
+    logger.warn('Unable to update subscriptions from notifier service provider', error)
+  }
```
jurajpiar, line 73: How about this @farinalvaro?

```suggestion
  const logError = (error) => logger.warn('Unable to update subscriptions from notifier service provider', error)
  
  const subscriptionsDTO: any[] = await svcProvider
  .getSubscriptions(consumerAddress, subsHashesToUpdate)
  .catch(logError)
  
  await Promise.all(subscriptionsDTO.map(({
    hash,
    status,
    paid,
    notificationBalance
  }) =>
    SubscriptionModel.update(
      { status, paid, notificationBalance },
      { where: { hash } }
    ).catch(logError)
  ))
 
```

jurajpiar (approved):

src/services/notifier/utils/updaterUtils.ts
itofarina: The problem with this one is that if the first one fails it keeps going and `subscriptionsDTO` will be `undefined`, and then it will crash “`subscriptionsDTO.map` cannot read property map of undefined”.

PR #724, opened 2021-07-02
fix(notifier): adds query to subscritptions

jurajpiar:
Resolves https://rsklabs.atlassian.net/browse/RMKT-793 by adding the orig query to the where clause


itofarina (approved):
👌 thanks for spotting it

0xartem (approved):
LGTM

PR #725, opened 2021-07-02
fix(notifier): fixes hashes grouping

itofarina:
Fixes hash grouping when fetching subscriptions for different consumers.

Error: 
image


vjgene (approved):
Good catch

PR #726, opened 2021-07-05
fix(notifier): adds params to query unique plan

jurajpiar:
Relates to https://rsklabs.atlassian.net/browse/RMKT-792.

What needed to be done is to follow the design change of the plan table (done as feature for url switching) when querying from subscription hook.


0xartem (approved):
good, thanks

PR #737, opened 2021-07-05
fix(notifier): includes expiration date while sync'ing subscriptions

vjgene:
Required for https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60?selectedIssue=RMKT-798

The expiration date could be different if the payment was made on a different date. This should be sync'd as well.


0xartem (approved):
looks good

PR #738, opened 2021-07-05
chore(backport): adds hotfix backporting

jurajpiar:


.github/workflows/backport.yml
```diff
@@ -0,0 +1,20 @@
+name: Backport
```
0xartem, line 1: i think the name should be 'Backport staging'

0xartem (changes requested):

.github/workflows/backport.yml
jurajpiar: Not really because this same file applies to all branches

0xartem: oh, okay

0xartem (approved):

0xartem (approved):

PR #740, opened 2021-07-07
feat(notifier): updates or creates subscriptions while fetching

itofarina:
Syncs cache subscriptions with notifier providers when fetching them
Resolves [RMKT-804](https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60?selectedIssue=RMKT-804)

_This PR should be squashed_

### DIFF
--- a/src/services/notifier/handlers/provider.ts
+++ b/src/services/notifier/handlers/provider.ts
@@ -2,17 +2,14 @@ import { Eth } from 'web3-eth'
 import { ProviderRegistered, SubscriptionCreated } from '@rsksmart/rif-marketplace-notifier/types/web3-v1-contracts/NotifierManager'
 
 import { loggingFactory } from '../../../logger'
-import { Handler, NotificationManagerEvents, SupportedServices } from '../../../definitions'
+import { Handler, NotificationManagerEvents } from '../../../definitions'
 import { wrapEvent } from '../../../utils'
 
 import { NotifierServices } from '../index'
 import ProviderModel from '../models/provider.model'
 import { updater } from '../update'
 import { NotifierSvcProvider } from '../notifierService/provider'
-import SubscriptionModel from '../models/subscription.model'
-import { getTokenSymbol } from '../../utils'
-import BigNumber from 'bignumber.js'
-import { deactivateDeletedPlansForProvider } from '../utils/updaterUtils'
+import { createSubscription, deactivateDeletedPlansForProvider } from '../utils/updaterUtils'
 
 const logger = loggingFactory('notifier:handler:provider')
 
@@ -55,46 +52,7 @@ export const handlers = {
     const [subscriptionDTO] = await notifierService.getSubscriptions(consumer, [hash])
 
     if (!subscriptionDTO) return
-
-    const {
-      currency: {
-        address: {
-          value: tokenAddress
-        }
-      },
-      price,
-      expirationDate,
-      id: subscriptionId,
-      paid,
-      status,
-      notificationBalance,
-      subscriptionPlanId,
-      previousSubscription: previousSubscriptionModel,
-      topics,
-      signature
-    } = subscriptionDTO
-
-    const tokenSymbol = getTokenSymbol(tokenAddress, SupportedServices.NOTIFIER).toLowerCase()
-    const previousSubscriptionHash = previousSubscriptionModel?.hash
-
-    const subscription = {
-      providerId: provider,
-      price: new BigNumber(price),
-      rateId: tokenSymbol,
-      expirationDate: new Date(expirationDate),
-      hash,
-      consumer,
-      subscriptionId,
-      paid,
-      status,
-      notificationBalance,
-      subscriptionPlanId,
-      previousSubscription: previousSubscriptionHash,
-      topics,
-      signature
-    }
-
-    await SubscriptionModel.create(subscription)
+    const subscription = await createSubscription(subscriptionDTO, provider)
 
     if (subscriptionService.emit) subscriptionService.emit('created', wrapEvent('SubscriptionCreated', subscription))
     logger.info(`Created new Subscription ${hash} by Consumer ${consumer} for Provider ${provider}`)
--- a/src/services/notifier/handlers/provider.ts
+++ b/src/services/notifier/handlers/provider.ts
@@ -9,7 +9,8 @@ import { NotifierServices } from '../index'
 import ProviderModel from '../models/provider.model'
 import { updater } from '../update'
 import { NotifierSvcProvider } from '../notifierService/provider'
-import { createSubscription, deactivateDeletedPlansForProvider } from '../utils/updaterUtils'
+import { buildSubscriptionFromDTO, deactivateDeletedPlansForProvider } from '../utils/updaterUtils'
+import SubscriptionModel from '../models/subscription.model'
 
 const logger = loggingFactory('notifier:handler:provider')
 
@@ -50,9 +51,10 @@ export const handlers = {
     const [host, port] = providerIns.url.split(/(?::)(\d*)$/, 2)
     const notifierService = new NotifierSvcProvider({ host, port })
     const [subscriptionDTO] = await notifierService.getSubscriptions(consumer, [hash])
+    const subscription = buildSubscriptionFromDTO(subscriptionDTO, provider)
 
     if (!subscriptionDTO) return
-    const subscription = await createSubscription(subscriptionDTO, provider)
+    await SubscriptionModel.create(subscription)
 
     if (subscriptionService.emit) subscriptionService.emit('created', wrapEvent('SubscriptionCreated', subscription))
     logger.info(`Created new Subscription ${hash} by Consumer ${consumer} for Provider ${provider}`)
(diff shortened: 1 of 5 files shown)


src/services/notifier/handlers/provider.ts
```diff
@@ -53,47 +51,9 @@ export const handlers = {
     const [host, port] = providerIns.url.split(/(?::)(\d*)$/, 2)
     const notifierService = new NotifierSvcProvider({ host, port })
     const [subscriptionDTO] = await notifierService.getSubscriptions(consumer, [hash])
+    const subscription = buildSubscriptionFromDTO(subscriptionDTO, provider)
 
     if (!subscriptionDTO) return
```
vjgene, line 56: It's probably better to move this line above buildSubscriptionFromDTO call

src/services/notifier/hooks/subscriptions.hook.ts
```diff
@@ -46,29 +46,25 @@ export default {
           ]
```
vjgene, line 44: we can remove the type type HashesByProviderAndConsumer = Record<string, Record<string, string[]>> which is not used anywhere

src/services/notifier/hooks/subscriptions.hook.ts
```diff
@@ -46,29 +46,25 @@ export default {
...
+        // updates DB fetching subscriptions directly from the notifier service
         const updateDataPromises: Promise<void>[] = []
-        Object.keys(hashesByProviderAndConsumer).forEach(providerUrl => {
-          const providerSubscriptions = hashesByProviderAndConsumer[providerUrl]
-          Object.keys(providerSubscriptions).forEach(consumerAddress => {
-            const hashes = providerSubscriptions[consumerAddress]
-            updateDataPromises.push(updateSubscriptionsBy(providerUrl, consumerAddress, hashes))
+        Object.keys(consumersByProviders).forEach(providerUrl => {
+          consumersByProviders[providerUrl].forEach(consumerAddress => {
+            updateDataPromises.push(updateSubscriptionsBy(providerUrl, consumerAddress))
```
jurajpiar: We don't need this any more, since we pull all the subscriptions. we can filter them post fetch instead, should we need to.

src/services/notifier/notifierService/provider.ts
```diff
@@ -45,14 +67,14 @@ export class NotifierSvcProvider extends ServiceProvider<NotifierResult<any>> im
...
-  async getSubscriptions (address: string, hashes: string[]) {
+  async getSubscriptions (address: string, hashes?: string[]) {
     const {
       success,
       message,
       code,
       data
     } = await this._fetch({
-      path: `/getSubscriptions/${hashes.toString()}`,
+      path: `/getSubscriptions/${hashes ? hashes.toString() : ''}`,
```
jurajpiar:
```suggestion
      path: `/getSubscriptions/${hashes?.toString() ?? ''}`,
```

src/services/notifier/utils/updaterUtils.ts
```diff
@@ -50,25 +54,85 @@ export async function deactivateDeletedPlansForProvider (provider: string, url:
...
-    ))
+    const subscriptionsDTO: Array<SubscriptionDTO> = await svcProvider.getSubscriptions(consumerAddress)
+
+    await Promise.all(
+      subscriptionsDTO.map(
+        (subscription) => findOrCreateSubscription(
+          subscription, providerUrl
+        ).catch(
+          error => logger.warn(`Unable to update or create subscription with hash ${subscription.hash} in the database`, error)
+        )
```
jurajpiar, line 141: Why would a simple `SubscriptionModel.findOrCreate` not suffice?
something like:
```suggestion
        (subscription) => SubscriptionModel
          .findOrCreate(
            buildSubscriptionFromDTO(subscriptionDTO, provider),
            { where: { hash: subscription.hash }}
          ).catch(error => logger.warn(`Unable to update or create subscription with hash ${subscription.hash} in the database`, error)
          )
        )
```

src/services/notifier/utils/updaterUtils.ts
```diff
@@ -50,25 +54,85 @@ export async function deactivateDeletedPlansForProvider (provider: string, url:
...
+    paid,
+    status,
+    notificationBalance,
+    subscriptionPlanId,
+    previousSubscription: previousSubscriptionHash,
+    topics,
+    signature
+  }
+
+  return subscription
```
jurajpiar: Here we need to do a couple of things: `subscriptionPlanId` needs to be set to cache-generated plan id, and we need to check for unsupported tokens just in case.  
```suggestion
  const {
    currency: {
      address: {
        value: tokenAddress
      }
    },
    price,
    provider: providerId,
    expirationDate,
    id: subscriptionId,
    subscriptionPlanId,
    previousSubscription: previousSubscriptionModel,
    userAddress: consumer,
    ...subscription
  } = subscriptionDTO
  const rateId = getTokenSymbol(tokenAddress, SupportedServices.NOTIFIER).toLowerCase()
  const plan = await PlanModel.findOne({
    where: {
      planId: subscriptionPlanId,
      planStatus: 'ACTIVE',
      providerId: providerId
    }
  })
  if (!(plan && rateId)) throw 'bad subscription dto'

  return {
    rateId,
    providerId,
    subscriptionId,
    subscriptionPlanId: plan.id,
    price: new BigNumber(price),
    expirationDate: new Date(expirationDate),
    previousSubscription: previousSubscriptionModel?.hash,
    ...subscription
  }
```

jurajpiar (changes requested):
Few things that caught my attention

src/services/notifier/hooks/subscriptions.hook.ts
```diff
@@ -46,29 +46,25 @@ export default {
           ]
         })
 
-        // provider -> {consumer} -> [hashes]
-        const hashesByProviderAndConsumer = accountSubscriptions.reduce(
-          (acc: HashesByProviderAndConsumer, current: SubscriptionModel) => {
-            const { consumer, hash, provider: { url } } = current
+        // providerUrl -> [consumerAddress]
+        const consumersByProviders = accountSubscriptions.reduce(
```
vjgene: If the tmpQuery has consumerAddress, then we don't need to group by consumers right? since it's already for a given consumer. we just need to find distinct provider url from the given set of subscriptions for the given consumer. So, what about having logic like
```

if (tmpQuery) {
  const providerList = new Set(accountSubscriptions.map(subscription=>subscription.url))
  providerList.forEach(prov=>updateSubscriptionsBy(prov, consumerAddress)
}
}
```

Also we should also consider a scenario where there are no subscriptions for a given provider in cache, but the notifier provider has some subscriptions created that are not in cache. Meaning it's a new provider from consumer perspective.

So would it be safe to do something 
```
const providers = await ProviderModel.findAll()
providers.forEach(p=>updateSubscriptionsBy(p.url, consumerAddress)
```

However this may have a performance impact if there are large number of providers.

vjgene (commented):
Looks good, few questions and suggestions

src/services/notifier/hooks/subscriptions.hook.ts
jurajpiar: I agree  with the first suggestion, for sure. The second one could be expensive in the long term, though if we are the only providers for the foreseeable future then it should be fine, too. What do you think @artem-brazhnikov?

0xartem: @jurajpiar @vjgene I think we can create a separate story for this and address if it becomes a priority in the future. @vjgene document it please in jira

0xartem (approved):

0xartem (changes requested):
waiting for reported discussions to be resolved

src/services/notifier/hooks/subscriptions.hook.ts
itofarina: I'm not sure I got the suggestion. We are not pulling all the subscriptions, only the ones related to consumers/providers already registered in cache.

src/services/notifier/hooks/subscriptions.hook.ts
vjgene: @artem-brazhnikov @jurajpiar created https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60?selectedIssue=RMKT-822

src/services/notifier/utils/updaterUtils.ts
itofarina: Because when finding it, we need to take action and also update `{ status, paid, notificationBalance, expirationDate }` fields

src/services/notifier/hooks/subscriptions.hook.ts
itofarina: Done with first suggestion, thanks

src/services/notifier/hooks/subscriptions.hook.ts
jurajpiar: well, yea I did mean all subscriptions for given consumer.

itofarina: For every provider we need to iterate over the involved consumers as to call `getSubscriptions` we need to provide the involved consumer. [Reference](https://github.com/rsksmart/rif-notifier#get-subscriptions)

src/services/notifier/utils/updaterUtils.ts
itofarina: Great suggestion, thanks

vjgene (approved):

0xartem (approved):

jurajpiar (approved):

PR #759, opened 2021-07-26
feat(notifier): adds ability to withdraw funds as a provider

itofarina:
## Features
- Listen and emits subscription funds withdrawn event
- Calculates and stores new column `withdrawableFunds` for each subscription when fetching data from notifier service (adds new migration)

## Known issues
As discussed, we could set `withdrawableFunds` to `BigNumber(0)` when withdraw event happens. The fact is that then it will get overwritten with notifier provider service data when fetching subscriptions. So, I'm not sure if it makes sense to add that operation, in any case, we can remove the last commit from this PR.

Related to [RMKT-839](https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60?selectedIssue=RMKT-839)

### DIFF
--- a/src/services/notifier/handlers/provider.ts
+++ b/src/services/notifier/handlers/provider.ts
@@ -1,5 +1,9 @@
 import { Eth } from 'web3-eth'
-import { ProviderRegistered, SubscriptionCreated } from '@rsksmart/rif-marketplace-notifier/types/web3-v1-contracts/NotifierManager'
+import {
+  ProviderRegistered,
+  SubscriptionCreated,
+  FundsWithdrawn
+} from '@rsksmart/rif-marketplace-notifier/types/web3-v1-contracts/NotifierManager'
 
 import { loggingFactory } from '../../../logger'
 import { Handler, NotificationManagerEvents } from '../../../definitions'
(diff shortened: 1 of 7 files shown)


0xartem (approved):
LGTM

src/services/notifier/handlers/provider.ts
```diff
@@ -77,6 +82,29 @@ export const handlers = {
...
+
+    const withdrawnEvent = {
+      provider,
+      hash,
+      amount,
+      token
+    }
+    try {
+      await SubscriptionModel.update(
+        { withdrawableFunds: new BigNumber(0) },
```
vjgene: I agree based on comments in description, this event is probably unnecessary and also it's assuming all funds withdrawn by setting to zero. So, it's probably better to rely on notifier since we are doing that anyway in findOrCreateSubscription

vjgene (approved):
Looks good. Just added comment to see if the event can be removed.

src/services/notifier/handlers/provider.ts
itofarina: Thanks, makes sense. I just removed the last commit.

PR #781, opened 2021-08-02
fix(devops): removes scripts copy to avoid build failure

rafaiovlabs:
* As "scripts" file doesn't exists,  we must erase this line, because it's producing building failures.


itofarina (approved):

0xartem (approved):

0xartem:
@rafaiovlabs the CI fails because of the commit message linting, please fix it

PR #815, opened 2021-09-09
fix(notifier): deactivates deprecated plans

jurajpiar:
In the plans update mechanism, it marks plans that are no longer present on the provider as inactive.

Resolving https://rsklabs.atlassian.net/browse/RMKT-915.

Added a unit test, but not tested against real data.


itofarina (approved):
LGTM - Please, just remove the unused `Op` reported by the bot

PR #824, opened 2021-09-09
fix(devops): removes only production ci

jurajpiar:
The build in testnet keeps failing due to type errors. These errors appear as a result of not installing the typing dependencies, which only exist in the devDependencies.


rafaiovlabs (approved):
LGTM

